PR #71, opened 2015-09-17
Use exit codes of 2 and 3

msabramo:
so that we can distinguish between the two different ways that
pycobertura can explicitly fail and even more importantly distinguish
these types of failures from exit code 1, which is what seems to be
returned if an exception occurs.

Currently, any kind of failure returns an exit code 1 so the caller
cannot distinguish between various cases.

Cc: @aconrad, @sudarkoff

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* return different exit codes depending on status:
+  0: all is good.
+  1: some exception occurred (likely a bug)
+  2: coverage worsened (implies 3)
+  3: not all changes are covered
+
 ## 0.7.3 (2015-07-23)
 
 * a non-zero exit code will be returned if not all changes have been
--- a/pycobertura/cli.py
+++ b/pycobertura/cli.py
@@ -15,6 +15,31 @@ reporters = {
 }
 
 
+class ExitCodes:
+    OK = 0
+    EXCEPTION = 1
+    COVERAGE_WORSENED = 2
+    NOT_ALL_CHANGES_COVERED = 3
+
+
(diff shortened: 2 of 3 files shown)


aconrad:
Thanks @msabramo!

So we'd have:
* `0`: all is good.
* `1`: some exception occurred (likely a bug)
* `2`: coverage worsened (implies `3`)
* `3`: not all changes are covered

Looks great. Can you update the CHANGES file as well as the tests?

msabramo:
Will do...

msabramo:
@aconrad: Updated tests and `CHANGES.md` and also did some refactoring that hopefully you like.

aconrad:
Thanks! Mind also testing the class `ExitCodes` since you've started using it for tests? Right now if I were to change the exit codes it wouldn't fail.

msabramo:
> Thanks! Mind also testing the class ExitCodes since you've started using it for tests? Right now if I were to change the exit codes it wouldn't fail.

Done.

msabramo:
I wonder in retrospect if I should've not used exit code 1 -- that way exit code 1 would signify that we're running an older version of pycobertura and hence, don't know what it means.

So we'd have:
* `0`: all is good.
* `1`: old version of pycobertura before it used different exit codes; we don't know why it failed unless we inspect output.
* `2`: some exception occurred (likely a bug)
* `3`: coverage worsened (implies `3`)
* `4`: not all changes are covered

aconrad:
Released v0.8.0.

I just saw your note after releasing. But since we're not 1.0 just yet, I don't really want to worry about old versions. I updates the CHANGES file to mention that it's a backwards incompatible change for those who care.

PR #72, opened 2015-11-07
Added coverage v4.0.x compatibility by changing class element lookups to use 'filename' instead of the 'name' attribute

JonSchuff:
**Description:**

Changed cobertura's coverage.xml parsing behavior to lookup class elements by filename attribute for compatibility with coverage v3.7.1 and versions >=4.0.0 and updated the tests accordingly.

**Rationale:**

In newer versions of the coverage library (> 3.7.1), the coverage.xml class element's 'name' attribute has changed from containing the filepath w/o the extension ( ex. "project/view/myview" ) to the base filename only ( ex. "myview.py" ). The package path context is now represented within the package's name attribute only, which means collisions during name-based lookups are bound to occur ("project/view/\__init\__.py" and "project/\__init\__.py" for instance), which will cause the tool to report inaccurate results. Using the filename attribute for identification and lookups is an easy win for compatibility across versions 3.7.1 and up.

This hasn't impacted everyone internally *yet* since our current pip index only contains coverage 3.7.1, implicitly locking everyone's coverage testing to that version. However, it will affect all projects that don't have coverage version locked since pytest-cov requires `coverage>=3.7.1` and devpi provides the newest version it can find (4.0.2).

Edit: here's an [example](https://gist.github.com/JonSchuff/78f4d4ad3bb16bb8bf3a) of the coverage.xml changes.

### DIFF
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -47,7 +47,7 @@ class Cobertura(object):
 
     @memoize
     def _get_element_by_class_name(self, class_name):
-        syntax = "./packages/package/classes/class[@name='%s'][1]" % class_name
+        syntax = "./packages/package/classes/class[@filename='%s'][1]" % class_name
         return self.xml.xpath(syntax)[0]
 
     @memoize
@@ -223,7 +223,7 @@ class Cobertura(object):
         """
         Return the list of available classes in the coverage report.
         """
-        return [el.attrib['name'] for el in self.xml.xpath("//class")]
+        return [el.attrib['filename'] for el in self.xml.xpath("//class")]
 
     def has_class(self, class_name):
         """
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -185,9 +185,9 @@ class Cobertura(object):
 
     def total_statements(self, class_filename=None):
         """
-        Return the total number of statements for the class file `class_filename`.
-        If `class_filename` is not given, return the total number of statements for
-        all class files.
+        Return the total number of statements for the class file
+        `class_filename`. If `class_filename` is not given, return the total
+        number of statements for all class files.
         """
         if class_filename is not None:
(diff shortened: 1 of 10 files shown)


aconrad:
Thanks! Your intent makes sense and the changes look good. I'll review this more thoroughly next week if that's okay.

pycobertura/cobertura.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, xml_path, base_path=None):
 
     @memoize
     def _get_element_by_class_name(self, class_name):
-        syntax = "./packages/package/classes/class[@name='%s'][1]" % class_name
+        syntax = "./packages/package/classes/class[@filename='%s'][1]" % class_name
```
aconrad: The tests fail because this is not PEP-8 compliant. Can you please wrap around the line like this?

```python
    @memoize
    def _get_element_by_class_name(self, class_name):
        syntax = "./packages/package/classes/class[@filename='%s'][1]" % (
            class_name
        )
        return self.xml.xpath(syntax)[0]
```

aconrad:
Can you please add the following entry to the `CHANGES.md` file under the `Unreleased` paragraph.

```
* The coverage report now displays the class's filename instead of the class's
  name, the latter being more subject to different interpretations by coverage
  tools. This change was done to support coverage.py versions 3.x and 4.x.
```

aconrad:
Additionally, we should change any reference of "class_name" to "class_filename" since we're fundamentally changing how we identify a class. So renaming methods, doc strings, ... Are you up for the challenge? :)

Maybe we should do this as part of another PR though (which I could do).

JonSchuff:
Sure, I'll make the requested changes- just wanted to make sure this was an acceptable fix before updating everything else :)

JonSchuff:
@aconrad ready for review

aconrad:
Thanks @JonSchuff

PR #77, opened 2016-09-16
Initialize status and reason

zoetetal:
In cases where you're calling the diff function but not referencing two different source trees, newly added lines can lead to situations where one line_status is None while the other one has a boolean value as expected. That ends up potentially never initializing status and reason.


aconrad:
Also, ideally, please add a test case that reproduces the issue.

CHANGES.md
```diff
@@ -11,6 +11,7 @@
 * Fix an `IOError` / `FileNotFound` error which happens when the same coverage report is provided twice
   to `pycobertura diff` (diff in degraded mode) but the first code base (`--source1`) is missing a file
   mentioned in the coverage report.
+* Fix a rare bug when diffing coverage xml where one file goes from zero lines to non-zero lines.
```
aconrad: please clean up the trailing whitespace at the end of this line

aconrad:
@zoetetal this looks great, thank you so much! I have a minor comment but other than that, it's good to go!

zoetetal:
OK thanks @aconrad just getting the final whitespace fix in. I forgot that as soon as we updated master on my fork the PR would update. Anyway if you remove the initialized status and reason and run tests you'll see the issue. Thanks for a nice library, we're using it during validation of pull requests.

aconrad:
Thanks @zoetetal, I merge-squashed. I will try to make a new release of pycobertura this week.

aconrad:
Released pycobertura v0.10.0, thanks again @zoetetal!

PR #84, opened 2018-04-12
Update TravisCI badge and link in README.md

z4y4ts:


aconrad:
Good catch and thank you, @z4y4ts!

PR #90, opened 2019-05-31
Fix Pytest4.x compatibility error

stanislavlevin:
According to pytest docs:

```
[pytest] section in setup.cfg files

Removed in version 4.0.

[pytest] sections in setup.cfg files should now be named [tool:pytest] to
avoid conflicts with other distutils commands.
```


aconrad:
Thanks for the fix @stanislavlevin!

PR #92, opened 2020-01-15
Added ability to works with non-utf8 sources

rburtnykx:
Currenlty pycobertura not able to work with sources other than utf-8. We faced with such issue. Maybe it is good to give ability to configure it?

### DIFF
--- a/pycobertura/filesystem.py
+++ b/pycobertura/filesystem.py
@@ -13,9 +13,11 @@ class FileSystem(object):
 
 
 class DirectoryFileSystem(FileSystem):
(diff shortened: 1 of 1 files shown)


graydh:
Why was this closed? I need the same feature ideally for the CLI. I'm willing to create a new PR/branch with this feature added but want to make sure there wasn't a difference solution found.

aconrad:
@graydh I don't think there was a discussion around this. Feel free to submit a PR if that's something that you think would be useful for Pycobertura.

PR #93, opened 2020-04-19
feat: a customizable html report

nilleb:
Let the HtmlReport caller
- specify a title for the report
- skip the sources detail (useful for big projects)
- replace the sources block with a message about why they are missing

The report you obtain looks like
Capture d’écran 2020-04-19 à 15 20 06

### DIFF
--- a/pycobertura/reporters.py
+++ b/pycobertura/reporters.py
@@ -95,6 +95,9 @@ class TextReporter(Reporter):
 
 class HtmlReporter(TextReporter):
     def __init__(self, *args, **kwargs):
+        self.title = kwargs.pop("title", "pycobertura report")
+        self.inline_sources = kwargs.pop("inline_sources", True)
+        self.sources_message = kwargs.pop('sources_message', None)
         super(HtmlReporter, self).__init__(*args, **kwargs)
 
     def get_source(self, filename):
@@ -110,15 +113,18 @@ class HtmlReporter(TextReporter):
             formatted_lines.append(formatted_row)
 
         sources = []
-        for filename in self.cobertura.files():
-            source = self.get_source(filename)
-            sources.append((filename, source))
+        if self.inline_sources:
+            for filename in self.cobertura.files():
+                source = self.get_source(filename)
+                sources.append((filename, source))
 
         template = env.get_template('html.jinja2')
         return template.render(
+            title=title,
             lines=formatted_lines[:-1],
             footer=formatted_lines[-1],
-            sources=sources
+            sources=sources,
+            sources_message=self.sources_message
         )
 
 
--- a/pycobertura/reporters.py
+++ b/pycobertura/reporters.py
@@ -123,7 +123,7 @@ class HtmlReporter(TextReporter):
 
         template = env.get_template('html.jinja2')
         return template.render(
-            title=title,
+            title=self.title,
             lines=formatted_lines[:-1],
             footer=formatted_lines[-1],
             sources=sources,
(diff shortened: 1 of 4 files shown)


pycobertura/templates/html.jinja2
```diff
@@ -13,6 +13,7 @@ pre {line-height: 1.3}
   </head>
   <body>
     <div class="container">
+      <h1> {{ title }} </h1>
```
aconrad, line 16: Whitespace.

```suggestion
      <h1>{{ title }}</h1>
```

pycobertura/reporters.py
```diff
@@ -95,6 +95,9 @@ def generate(self):
 
 class HtmlReporter(TextReporter):
     def __init__(self, *args, **kwargs):
+        self.title = kwargs.pop("title", "pycobertura report")
+        self.inline_sources = kwargs.pop("inline_sources", True)
```
aconrad, line 99: Can we rename this to:

```suggestion
        self.render_file_sources = kwargs.pop("render_file_sources", True)
```

...?

pycobertura/templates/html.jinja2
```diff
@@ -44,11 +51,15 @@ pre {line-height: 1.3}
...
         </tfoot>
       </table>
+{% if sources %}
 {%- from 'macro.source.jinja2' import render_source -%}
 {%- for filename, source in sources %}
 <h4 id="{{ filename }}">{{ filename }}</h4>
 {{ render_source(source) }}
 {%- endfor %}
+{% else %}
+    {{ sources_message }}
```
aconrad, line 61: Should this be wrapped in a `<p>` tag or something? Also, can we make the indentation in HTML files 2-space?

```suggestion
  <p>{{ sources_message }}</p>
```

pycobertura/reporters.py
```diff
@@ -95,6 +95,9 @@ def generate(self):
...
+        self.sources_message = kwargs.pop('sources_message', None)
```
aconrad, line 100: Can we have a fallback message in case no message is provided to prevent confusion? Also, can we also rename `no_file_sources_message`.

```suggestion
        self.no_file_sources_message = kwargs.pop(
            "no_file_sources_message", "Rendering of source files was disabled."
        )
```

But for your use-case, do you really need to customize the message? Wouldn't a generic message "Rendering of source files was disabled" satisfy your needs? Why do you need to customize it in the first place?

pycobertura/templates/html.jinja2
```diff
@@ -26,7 +27,13 @@ pre {line-height: 1.3}
         <tbody>
 {%- for line in lines %}
           <tr>
-            <td><a href="#{{ line.filename }}">{{ line.filename }}</a></td>
+            <td>
+            {% if sources %}
+                <a href="#{{ line.filename }}">{{ line.filename }}</a>
+            {% else %}
+                {{ line.filename }}
+            {% endif %}
```
aconrad, line 35: Whitespace.

```suggestion
            {% if sources %}
              <a href="#{{ line.filename }}">{{ line.filename }}</a>
            {% else %}
              {{ line.filename }}
            {% endif %}
```

nilleb:
The `no_file_sources_message` is useful to pycobertura consumers which have an alternate way to produce the report. In the example screenshot, the report has been produced on a CI environment. The message in the sample is pretty useless, but I imagined to modify it in order to explain the visitor how to rebuild the same report on his developer machine, including the sources, if he has some doubts about how he could improve his code coverage.

aconrad:
This PR looks good to me, thanks for the updates @nilleb!

Can we please add some tests in:
https://github.com/aconrad/pycobertura/blob/master/tests/test_reporters.py

I have to revisit the CI integration, I'm not sure why Travis isn't working.

Also, please remember to update the CHANGES file and add yourself as a contributor. You might run into a merge conflict when your other PR gets merged but it's a reminder that we need to add a changelog entry.

aconrad:
@nilleb thanks for your work on this! Your contribution is much appreciated!

PR #94, opened 2020-04-19
feat: a gitfilesystem

nilleb:
The consumer of the VersionedCobertura has the possibility to inspect a coverage report from the past along with the corresponding sources.

Usage:
```python
VersionedCobertura(report, source=source, commit_id=commit_id)
```
Where `report` is the Cobertura report, `source` is the folder where the Git repository can be found and `commit` is the commit ID to which the report is associated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * Let the caller customize the appearance of the HTML report providing a `title`, omitting the rendering of sources by means of the boolean `render_file_sources` and providing an helpful message to the end-users (in place of the sources) by means of the `no_file_sources_message` parameter. Contributed by @nilleb.
-* Add a GitFilesystem, allowing the consumer to retrieve the sources from a git commit.
+* Add a `GitFilesystem` to allow pycobertura to access source files at different revisions from a git repository. Thanks @nilleb.
 * [BREAKING] Change the signature of the Cobertura object in order to accept a filesystem.
 
 ## 0.10.5 (2018-12-11)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 * Let the caller customize the appearance of the HTML report providing a `title`, omitting the rendering of sources by means of the boolean `render_file_sources` and providing an helpful message to the end-users (in place of the sources) by means of the `no_file_sources_message` parameter. Contributed by @nilleb.
+* Add a GitFilesystem, allowing the consumer to retrieve the sources from a git commit.
+* [BREAKING] Change the signature of the Cobertura object in order to accept a filesystem.
 
 ## 0.10.5 (2018-12-11)
 * Use a different `memoize()` implementation so that cached objects can be freed/garbage collected and prevent from running out of memory when processing a lot of cobertura files. Thanks @kannaiah
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -431,10 +431,10 @@ class CoberturaDiff(object):
 
 
 class VersionedCobertura(Cobertura):
-    def __init__(self, report, source=None, commit_id=None):
+    def __init__(self, report, source=None, ref=None):
         super().__init__(report, source=source)
         if source is None:
             if isinstance(report, str):
                 # get the directory in which the coverage file lives
                 source = os.path.dirname(report)
-        self.filesystem = GitFileSystem(source, commit_id=commit_id)
+        self.filesystem = GitFileSystem(source, ref)
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -47,14 +47,12 @@ class Cobertura(object):
         module in order to discover more about filesystems.
         """
         self.xml = ET.parse(report).getroot()
-        self._report = report if isinstance(report, basestring) else None
-        self._filesystem = filesystem
+        report = report if isinstance(report, basestring) else None
 
-    @property
-    def filesystem(self):
-        if not self._filesystem:
-            self._filesystem = filesystem_factory(self._report)
-        return self._filesystem
+        if filesystem:
+            self.filesystem = filesystem
+        else:
+            self.filesystem = filesystem_factory(report)
 
     @memoize
     def _get_class_element_by_filename(self, filename):
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -1,14 +1,8 @@
 import lxml.etree as ET
-import os
-import zipfile
 
 from collections import namedtuple
 
-from pycobertura.filesystem import (
-    DirectoryFileSystem,
-    ZipFileSystem,
-    GitFileSystem
-)
+from pycobertura.filesystem import filesystem_factory
 
 
 from pycobertura.utils import (
@@ -42,37 +36,25 @@ class Cobertura(object):
     """
     An XML Cobertura parser.
     """
-    def __init__(self, report, source=None, source_prefix=None):
+    def __init__(self, report, filesystem=None):
         """
         Initialize a Cobertura report given a coverage report `report`. It can
         be either a file object or the path to an XML file that is in the
         Cobertura format.
 
-        The optional argument `source` is the location of the source code
-        provided as a directory path or a file object zip archive containing
-        the source code.
-
-        The optional argument `source_prefix` will be used to lookup source
-        files if a zip archive is provided and will be prepended to filenames
-        found in the coverage report.
+        The optional argument `filesystem` describes how to retrieve the source
+        files referenced in the report. Please check the pycobertura.filesystem
+        module in order to discover more about filesystems.
         """
         self.xml = ET.parse(report).getroot()
+        self._report = report if isinstance(report, basestring) else None
+        self._filesystem = filesystem
 
-        if source is None:
-            if isinstance(report, basestring):
-                # get the directory in which the coverage file lives
-                source = os.path.dirname(report)
-            self.filesystem = DirectoryFileSystem(
-                source, source_prefix=source_prefix
-            )
-        elif zipfile.is_zipfile(source):
-            self.filesystem = ZipFileSystem(
-                source, source_prefix=source_prefix
-            )
-        else:
-            self.filesystem = DirectoryFileSystem(
-                source, source_prefix=source_prefix
-            )
+    @property
+    def filesystem(self):
+        if not self._filesystem:
+            self._filesystem = filesystem_factory(self._report)
+        return self._filesystem
 
     @memoize
     def _get_class_element_by_filename(self, filename):
@@ -428,13 +410,3 @@ class CoberturaDiff(object):
         lines = self.file_source(filename)
         hunks = hunkify_lines(lines)
         return hunks
-
-
-class VersionedCobertura(Cobertura):
-    def __init__(self, report, source=None, ref=None):
-        super().__init__(report, source=source)
-        if source is None:
-            if isinstance(report, str):
-                # get the directory in which the coverage file lives
-                source = os.path.dirname(report)
-        self.filesystem = GitFileSystem(source, ref)
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -7,6 +7,7 @@ from collections import namedtuple
 from pycobertura.filesystem import (
     DirectoryFileSystem,
     ZipFileSystem,
+    GitFileSystem
 )
 
 
@@ -427,3 +428,13 @@ class CoberturaDiff(object):
         lines = self.file_source(filename)
         hunks = hunkify_lines(lines)
         return hunks
+
+
+class VersionedCobertura(Cobertura):
+    def __init__(self, report, source=None, commit_id=None):
+        super().__init__(report, source=source)
+        if source is None:
+            if isinstance(report, str):
+                # get the directory in which the coverage file lives
+                source = os.path.dirname(report)
+        self.filesystem = GitFileSystem(source, commit_id=commit_id)
--- a/pycobertura/filesystem.py
+++ b/pycobertura/filesystem.py
@@ -2,6 +2,7 @@ import codecs
 import os
 import zipfile
 
+from pycobertura.utils import get_output
 from contextlib import contextmanager
 
 
@@ -68,3 +69,52 @@ class ZipFileSystem(FileSystem):
             f.close()
         except KeyError:
             raise self.FileNotFound(filename)
+
+
+def get_root_path(repository_folder):
+    command = "git rev-parse --show-toplevel"
+    return get_output(command, working_folder=repository_folder).rstrip()
+
+
+class GitFileSystem(FileSystem):
+    def __init__(self, repo_folder, commit_id=None):
+        self.repository = repo_folder
+        self.commit_id = commit_id
+        self.repository_root = get_root_path(repo_folder)
+        self.prefix = self.repository.replace(self.repository_root, "").lstrip("/")
+
+    def real_filename(self, filename):
+        prefix = "{}/".format(self.prefix) if self.prefix else ""
+        return "{p.commit_id}:{prefix}{filename}".format(
+            prefix=prefix, p=self, filename=filename
+        )
+
+    def has_file(self, filename):
+        command = "git --no-pager show {}".format(self.real_filename(filename))
+        return_code = subprocess.call(
+            command,
+            cwd=self.repository,
+            shell=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+        return not bool(return_code)
+
+    @contextmanager
+    def open(self, filename):
+        """
+        Yield a file-like object for file `filename`.
+
+        This function is a context manager.
+        """
+        filename = self.real_filename(filename)
+
+        try:
+            output = get_output(
+                "git --no-pager show {}".format(filename), self.repository
+            )
+        except Exception:
+            raise self.FileNotFound(filename)
+
+        yield io.StringIO(output)
+
--- a/pycobertura/filesystem.py
+++ b/pycobertura/filesystem.py
@@ -1,8 +1,10 @@
 import codecs
 import os
+import io
 import zipfile
+import subprocess
+import shlex
 
-from pycobertura.utils import get_output
 from contextlib import contextmanager
 
 
@@ -71,22 +73,17 @@ class ZipFileSystem(FileSystem):
             raise self.FileNotFound(filename)
 
 
-def get_root_path(repository_folder):
-    command = "git rev-parse --show-toplevel"
-    return get_output(command, working_folder=repository_folder).rstrip()
-
-
 class GitFileSystem(FileSystem):
-    def __init__(self, repo_folder, commit_id=None):
+    def __init__(self, repo_folder, ref):
         self.repository = repo_folder
-        self.commit_id = commit_id
-        self.repository_root = get_root_path(repo_folder)
+        self.ref = ref
+        self.repository_root = self._get_root_path(repo_folder)
         self.prefix = self.repository.replace(self.repository_root, "").lstrip("/")
 
     def real_filename(self, filename):
         prefix = "{}/".format(self.prefix) if self.prefix else ""
-        return "{p.commit_id}:{prefix}{filename}".format(
-            prefix=prefix, p=self, filename=filename
+        return "{ref}:{prefix}{filename}".format(
+            prefix=prefix, ref=self.ref, filename=filename
         )
 
     def has_file(self, filename):
@@ -100,6 +97,11 @@ class GitFileSystem(FileSystem):
         )
         return not bool(return_code)
 
+    def _get_root_path(self, repository_folder):
+        command = "git rev-parse --show-toplevel"
+        output = subprocess.check_output(shlex.split(command), cwd=repository_folder)
+        return output.decode("utf-8").rstrip()
+
     @contextmanager
     def open(self, filename):
         """
@@ -109,12 +111,12 @@ class GitFileSystem(FileSystem):
         """
         filename = self.real_filename(filename)
 
+        command = "git --no-pager show {}".format(filename)
+
         try:
-            output = get_output(
-                "git --no-pager show {}".format(filename), self.repository
-            )
-        except Exception:
+            output = subprocess.check_output(shlex.split(command), cwd=self.repository)
+        except OSError:
             raise self.FileNotFound(filename)
 
+        output = output.decode("utf-8").rstrip()
         yield io.StringIO(output)
-
--- a/pycobertura/filesystem.py
+++ b/pycobertura/filesystem.py
@@ -42,7 +42,7 @@ class DirectoryFileSystem(FileSystem):
         if not os.path.exists(filename):
             raise self.FileNotFound(filename)
 
-        with codecs.open(filename, encoding='utf-8') as f:
+        with codecs.open(filename, encoding="utf-8") as f:
             yield f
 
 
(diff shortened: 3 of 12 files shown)


pycobertura/filesystem.py
```diff
@@ -68,3 +69,52 @@ def open(self, filename):
...
+        self.repository = repo_folder
+        self.commit_id = commit_id
+        self.repository_root = get_root_path(repo_folder)
+        self.prefix = self.repository.replace(self.repository_root, "").lstrip("/")
+
+    def real_filename(self, filename):
+        prefix = "{}/".format(self.prefix) if self.prefix else ""
+        return "{p.commit_id}:{prefix}{filename}".format(
+            prefix=prefix, p=self, filename=filename
+        )
```
aconrad, line 90: Can we instead do:

```suggestion
        return "{commit_id}:{prefix}{filename}".format(
            prefix=prefix, commit_id=self.commit_id, filename=filename
        )
```

... ? Would that work?

pycobertura/filesystem.py
```diff
@@ -68,3 +69,52 @@ def open(self, filename):
...
             raise self.FileNotFound(filename)
+
+
+def get_root_path(repository_folder):
+    command = "git rev-parse --show-toplevel"
+    return get_output(command, working_folder=repository_folder).rstrip()
+
+
+class GitFileSystem(FileSystem):
+    def __init__(self, repo_folder, commit_id=None):
```
aconrad, line 80: Should we make `commit_id` required? It seems that we're using it unconditionally.

```suggestion
    def __init__(self, repo_folder, commit_id):
```

pycobertura/filesystem.py
```diff
@@ -68,3 +69,52 @@ def open(self, filename):
...
+        self.repository = repo_folder
+        self.commit_id = commit_id
```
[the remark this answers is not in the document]
nilleb: Of course! _Bien vu_ ;-)

pycobertura/filesystem.py
```diff
@@ -68,3 +69,52 @@ def open(self, filename):
             f.close()
         except KeyError:
             raise self.FileNotFound(filename)
+
+
+def get_root_path(repository_folder):
```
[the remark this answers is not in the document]
nilleb: Sorry for having just copied and pasted some code from elsewhere. I agree with what you say, it totally makes sense in this organization of code.

pycobertura/utils.py
```diff
@@ -208,3 +210,15 @@ def hunkify_lines(lines, context=3):
...
+def get_output(command, working_folder=None):
+    logging.debug("Executing %s in %s", command, working_folder)
+
+    try:
+        output = subprocess.check_output(shlex.split(command), cwd=working_folder)
+        return output.decode("utf-8")
+    except OSError:
+        logging.error("Command being executed: {}".format(command))
+        raise
+
```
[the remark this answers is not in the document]
nilleb: I have noticed that in both my pull requests I left trailing whitespaces.

Would you mind if I use black to format sources? I gave it a try on your code (that is very nicely formatted), and if you were OK to switch to a line length of 88 characters, it would, I think, fit your needs.

pycobertura/filesystem.py
nilleb: Yeah, of course. When I first wrote this I was playing with format strings, and unfortunately this is still in the code.

pycobertura/filesystem.py
nilleb: 👍

pycobertura/utils.py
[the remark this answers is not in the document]
aconrad: I've never used black before. Also the homepage doesn't make it clear to what it covers. It seems like it would work like "prettier", no? Does it work only for Python?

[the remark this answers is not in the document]
aconrad: If we were to use Black, I'd rather do this in a separate PR if you don't mind.

nilleb:
I have also squeezed the `get_output` trying to rationalize the use of `subprocess` in `filesystem.py`.

Concerning the unit tests: of course, sorry. I am going to provide a distinct commit.
I have just seen that some unit tests are failing on python 27. Would you consider dropping unsupported versions of python from tox.ini? :-)

Concerning the trailing whitespaces: I have noticed that in both my pull requests I left some of them.
Would you mind if I use black to format sources? I gave it a try on your code (that is very nicely formatted), and if you were OK to switch to a line length of 88 characters, it would, I think, fit your needs. I will try to provide this in a distinct commit, so that if you are not OK I'll squeeze it out.

And, last, concerning the design patterns, yeah, I totally agree. The Cobertura __init__ is a factory method of a filesystem. The choice you have made was the best one, because simplicity is at stake. For the future, a constructor with a FileSystem argument instead of the existing ones, as you have suggested, could be the best one to improve code readability and single responsibility principle.

If you are OK to major bump, I'll propose a rework following your suggestions.

aconrad:
Happy to drop unsupported versions of python. We might as well do that if we're going to do a major version bump.

In addition, we should check if the readme needs to be updated accordingly with the changes you are about to make. And add the changes in the change log file, under "unreleased".

tests/test_filesystem.py
```diff
@@ -96,3 +96,52 @@ def test_filesystem_zip__with_source_prefix():
...
+    with patch.object(fsm, "subprocess") as subprocess_mock:
+        subprocess_mock.check_output = MagicMock(return_value=b"<file-content>")
+
+        fs = GitFileSystem(folder, branch)
+        git_filename = fs.real_filename(filename)
+
+        with fs.open(filename) as f:
+            assert hasattr(f, 'read')
+
+        assert git_filename.startswith(branch)
```
aconrad, line 116: Is it possible to assert for the whole filename just to be sure it contains everything necessary? For example:

```suggestion
        expected_git_filename = "master:tests/dummy/test-file"
        assert git_filename == expected_git_filename 
```

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 skipsdist = True
-envlist = py26, py27, py34, pep8
+envlist = py34, py37, py38, pep8
```
aconrad, line 3: Do we want to support py35 and/or py36? It has been a few years since I've worked on Python professionally so please excuse me if I'm a little rusty on my understanding of the Python ecosystem.

pycobertura/filesystem.py
```diff
@@ -68,3 +71,52 @@ def open(self, filename):
...
+        Yield a file-like object for file `filename`.
+
+        This function is a context manager.
+        """
+        filename = self.real_filename(filename)
+
+        command = "git --no-pager show {}".format(filename)
+
+        try:
+            output = subprocess.check_output(shlex.split(command), cwd=self.repository)
```
aconrad, line 117: Yay for using `subprocess`!

pycobertura/filesystem.py
```diff
@@ -68,3 +71,52 @@ def open(self, filename):
...
+        except (OSError, subprocess.CalledProcessError):
```
aconrad, line 118: Do we still need `OSError` here now that we use `subprocess`? In what circumstances would we expect `OSError` to get raised?

pycobertura/filesystem.py
```diff
@@ -68,3 +71,52 @@ def open(self, filename):
...
+            shell=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+        return not bool(return_code)
+
+    def _get_root_path(self, repository_folder):
+        command = "git rev-parse --show-toplevel"
+        output = subprocess.check_output(shlex.split(command), cwd=repository_folder)
+        return output.decode("utf-8").rstrip()
```
aconrad, line 103: Yay, thanks for making this method scoped to the class!

tests/test_filesystem.py
```diff
@@ -96,3 +96,52 @@ def test_filesystem_zip__with_source_prefix():
...
+    import pycobertura.filesystem as fsm
+    from pycobertura.filesystem import GitFileSystem
+
+    branch, folder, filename = "master", "tests/dummy", "test-file"
+
+    with patch.object(fsm, "subprocess") as subprocess_mock:
+        subprocess_mock.check_output = MagicMock(return_value=b"<file-content>")
+
+        fs = GitFileSystem(folder, branch)
+        git_filename = fs.real_filename(filename)
```
aconrad, line 111: Can we move this line closer to where it is being used a few lines below?

tests/test_filesystem.py
```diff
@@ -96,3 +96,52 @@ def test_filesystem_zip__with_source_prefix():
...
+
+        assert git_filename.startswith(branch)
+        expected_command = ["git", "--no-pager", "show", git_filename]
+        subprocess_mock.check_output.assert_called_with(expected_command, cwd=folder)
+
+
+def test_filesystem_git_integration():
+    from pycobertura.filesystem import GitFileSystem
+
+    fs = GitFileSystem(".", "d1fe88da6b18340762b24bb1f89067a3439c4041")
```
aconrad, line 124: Can we add a little explicitness as to what this commit is? Without a note, it leaves the reader wondering where that commit comes from.

```suggestion
    first_pycobertura_commit = "d1fe88da6b18340762b24bb1f89067a3439c4041"
    fs = GitFileSystem(".", first_pycobertura_commit)
```

Alternatively, we could also make it a CONSTANT at the top of this module and reuse the constant throughout this file:

```python
FIRST_PYCOBERTURA_COMMIT_SHA = "d1fe88da6b18340762b24bb1f89067a3439c4041"
```

... and then

```python
fs = GitFileSystem(".", FIRST_PYCOBERTURA_COMMIT_SHA)
```

What do you think?

tests/test_filesystem.py
```diff
@@ -96,3 +96,52 @@ def test_filesystem_zip__with_source_prefix():
...
+
+    for source_file in source_files:
+        with fs.open(source_file) as f:
+            assert hasattr(f, 'read')
+
+
+def test_filesystem_git_integration__not_found():
+    from pycobertura.filesystem import GitFileSystem
+
+    fs = GitFileSystem(".", "d1fe88da6b18340762b24bb1f89067a3439c4041")
```
aconrad, line 139: Same here.

```suggestion
    first_pycobertura_commit = "d1fe88da6b18340762b24bb1f89067a3439c4041"
    fs = GitFileSystem(".", first_pycobertura_commit)
```

You could also make it a CONSTANT at the top of this module:

```python
FIRST_PYCOBERTURA_COMMIT_SHA = "d1fe88da6b18340762b24bb1f89067a3439c4041"
```

... and then

```python
fs = GitFileSystem(".", FIRST_PYCOBERTURA_COMMIT_SHA)
```

tests/test_filesystem.py
```diff
@@ -96,3 +96,52 @@ def test_filesystem_zip__with_source_prefix():
...
+        expected_command = ["git", "--no-pager", "show", git_filename]
+        subprocess_mock.check_output.assert_called_with(expected_command, cwd=folder)
+
+
+def test_filesystem_git_integration():
+    from pycobertura.filesystem import GitFileSystem
+
+    fs = GitFileSystem(".", "d1fe88da6b18340762b24bb1f89067a3439c4041")
+
+    source_files = [
```
aconrad, line 126: Non-blocking: Can we add a small comment here to understand why we picked these files to test on?

```suggestion
    # Files included in pycobertura's first commit.
    source_files = [
```

pycobertura/filesystem.py
aconrad, line 117: Non-blocking: Is it possible that `shlex.split()` fails here? Should we move it outside of the try/except block so we reduce the scope of what is being executed within try/except?

```suggestion
        command = "git --no-pager show {}".format(filename)
        command = shlex.split(command)

        try:
            output = subprocess.check_output(command, cwd=self.repository)
```

pycobertura/filesystem.py
```diff
@@ -68,3 +71,52 @@ def open(self, filename):
...
+    def real_filename(self, filename):
+        prefix = "{}/".format(self.prefix) if self.prefix else ""
+        return "{ref}:{prefix}{filename}".format(
+            prefix=prefix, ref=self.ref, filename=filename
+        )
+
+    def has_file(self, filename):
+        command = "git --no-pager show {}".format(self.real_filename(filename))
+        return_code = subprocess.call(
+            command,
```
aconrad, line 92: Should we be using `shlex` here, too?

pycobertura/filesystem.py
```diff
@@ -68,3 +71,52 @@ def open(self, filename):
...
         except KeyError:
             raise self.FileNotFound(filename)
+
+
+class GitFileSystem(FileSystem):
+    def __init__(self, repo_folder, ref):
+        self.repository = repo_folder
+        self.ref = ref
+        self.repository_root = self._get_root_path(repo_folder)
+        self.prefix = self.repository.replace(self.repository_root, "").lstrip("/")
```
aconrad, line 81: Can you elaborate on how `prefix` is used in the context of `GitFileSystem`? It's not immediately obvious by reading the code.

tox.ini
aconrad: Also, I'm seeing here that Python 3.4 has reached end-of-life. Should we remove it from tox as well? https://www.python.org/downloads/release/python-340/

pycobertura/filesystem.py
nilleb: Well, according to https://docs.python.org/2/library/subprocess.html#exceptions
we should catch OSError (on systems where git is not available) and CalledProcessError (if the executable returns a non-zero error code).
I think this is a good idea to keep both.

aconrad: Oh great, thanks for looking into it!

nilleb:
I have addressed your latest review comments, but I have not yet changed the unreleased section of the changelog.
This because I am trying to add a last commit that will change the signature of the Cobertura constructor, as you suggested in https://github.com/aconrad/pycobertura/pull/94#pullrequestreview-396054566.
But if you prefer to keep the breaking changes for later, I'll just update the changelog.

pycobertura/cobertura.py
```diff
@@ -41,37 +36,25 @@ class Cobertura(object):
...
-            )
-        else:
-            self.filesystem = DirectoryFileSystem(
-                source, source_prefix=source_prefix
-            )
+    @property
+    def filesystem(self):
+        if not self._filesystem:
+            self._filesystem = filesystem_factory(self._report)
+        return self._filesystem
```
aconrad, line 57: What is the reason for adding this getter? Should we change the references in this file to access `self._filesystem` instead?

pycobertura/filesystem.py
```diff
@@ -68,3 +71,106 @@ def open(self, filename):
...
+            source, source_prefix=source_prefix
+        )
+
+    if os.path.isfile(source):
+        source = os.path.dirname(source)
+
+    if ref:
+        return GitFileSystem(source, ref)
+    else:
+        return DirectoryFileSystem(source, source_prefix=source_prefix)
```
aconrad, line 176: We can avoid the else-statement here since we use a guard-clause above.

```suggestion

    return DirectoryFileSystem(source, source_prefix=source_prefix)
```

pycobertura/filesystem.py
```diff
@@ -68,3 +71,106 @@ def open(self, filename):
...
+            # get the directory in which the coverage file lives
+            source = os.path.dirname(report)
+
+    if zipfile.is_zipfile(source):
+        return ZipFileSystem(
+            source, source_prefix=source_prefix
+        )
+
+    if os.path.isfile(source):
+        source = os.path.dirname(source)
```
aconrad, line 171: The keyword argument `source` is expected to be a directory according to the documentation, so I'm not sure when `source` could be a file path. In what scenario would this if-statement be true?

tests/test_filesystem.py
```diff
@@ -96,3 +99,55 @@ def test_filesystem_zip__with_source_prefix():
...
+def test_filesystem_git():
+    import pycobertura.filesystem as fsm
+    from pycobertura.filesystem import GitFileSystem
+
+    branch, folder, filename = "master", "tests/dummy", "test-file"
+
+    with patch.object(fsm, "subprocess") as subprocess_mock:
+        subprocess_mock.check_output = MagicMock(return_value=b"<file-content>")
+
+        fs = GitFileSystem(folder, branch)
```
aconrad, line 113: Non-blocking: could we reuse the import `fsm` and avoid the import `from pycobertura.filesystem import GitFileSystem` above?

```suggestion
        fs = fsm.GitFileSystem(folder, branch)
```

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Add a GitFilesystem, allowing the consumer to retrieve the sources from a git commit.
```
aconrad, line 5: Let's get you some credit for the work you have done here! 🎉

```suggestion
* Add a `GitFilesystem` to allow pycobertura to access source files at different revisions from a git repository. Thanks @nilleb.
```

pycobertura/cobertura.py
nilleb: Because when the Cobertura instance has been built with `Cobertura(xml)` there is no filesystem at all, and the two methods `file_source` and `source_lines` could throw. This property could be simply replaced with a 
`self.filesystem = filesystem if filesystem else filesystem_factory(report)` at line 51.
Or just dropped if you want your consumers to get exceptions when no filesystem has been passed at all.

aconrad: Gotcha.

Let's remove it as I would rather not expose this attribute as part of the public API. In hindsight, I should've made it internal/private in the first place. Also, now that the filesystem is being passed by the caller, they already have access to it should if they need the instance.

I would rather let it die or throw an exception that says that the Cobertura instance wasn't configured with a filesystem to be friendlier (but you don't have to do it, I could add that in a subsequent PR).

nilleb:
Given the times we live in, I hope that you are doing well. I just wanted to check how did you go. Have a good day!
Ivo

aconrad:
Hey @nilleb, thanks for checking in! I'm all good. I have a lot going on right now (new job search due to covid-19) which is why I've been a little slower to respond.

TLDR: I think your PR looks great! I'm ready to merge if we can bring up the coverage if `pycobertura/filesystem.py` back to 100% coverage. Can we do that?

_Below is a brain dump of some early investigation as to why the coverage is so off between py37 and py38._

---------------------

I ran your PR locally but I'm seeing the coverage report being really off. And I'm not sure why but it seems to only affect Python 3.8, prior versions of it work fine...

Here is the coverage report with py37 (against this PR):

```
===================================== warnings summary =====================================
.tox/py37/lib/python3.7/site-packages/_pytest/junitxml.py:417
  /mnt/c/Users/Alex/Documents/work/pycobertura/.tox/py37/lib/python3.7/site-packages/_pytest/junitxml.py:417: PytestDeprecationWarning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
  Add 'junit_family=xunit1' to your pytest.ini file to keep the current format in future versions of pytest and silence this warning.
    _issue_warning_captured(deprecated.JUNIT_XML_DEFAULT_FAMILY, config.hook, 2)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
-------- generated xml file: /mnt/c/Users/Alex/Documents/work/pycobertura/junit.xml --------

----------- coverage: platform linux, python 3.7.5-final-0 -----------
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
pycobertura/__init__.py                 2      0   100%
pycobertura/cli.py                     64      0   100%
pycobertura/cobertura.py              190      0   100%
pycobertura/filesystem.py              88      7    92%   94-103, 112-113, 171
pycobertura/reporters.py              164      0   100%
pycobertura/templates/__init__.py       0      0   100%
pycobertura/templates/filters.py        8      0   100%
pycobertura/utils.py                  111      1    99%   31
-----------------------------------------------------------------
TOTAL                                 627      8    99%
Coverage XML written to file coverage.xml

============================== 101 passed, 1 warning in 4.10s ==============================
_________________________________________ summary __________________________________________
  py37: commands succeeded
  congratulations :)
```

(NOTE: do you think you can get `pycobertura/filesystem.py` back to 100% coverage?)

... and here is the one against py38:

```
===================================== warnings summary =====================================
.tox/py38/lib/python3.8/site-packages/_pytest/junitxml.py:417
  /mnt/c/Users/Alex/Documents/work/pycobertura/.tox/py38/lib/python3.8/site-packages/_pytest/junitxml.py:417: PytestDeprecationWarning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
  Add 'junit_family=xunit1' to your pytest.ini file to keep the current format in future versions of pytest and silence this warning.
    _issue_warning_captured(deprecated.JUNIT_XML_DEFAULT_FAMILY, config.hook, 2)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
-------- generated xml file: /mnt/c/Users/Alex/Documents/work/pycobertura/junit.xml --------

----------- coverage: platform linux, python 3.8.0-final-0 -----------
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
pycobertura/__init__.py                 2      0   100%
pycobertura/cli.py                    109     81    26%   268-3726
pycobertura/cobertura.py              203     35    83%   315, 416-911
pycobertura/filesystem.py             103     26    75%   94-96, 339-886
pycobertura/reporters.py              207     99    52%   276, 289, 299, 315, 318, 320-1314
pycobertura/templates/__init__.py       0      0   100%
pycobertura/templates/filters.py       13      8    38%   261-536
pycobertura/utils.py                  111      1    99%   31
-----------------------------------------------------------------
TOTAL                                 748    250    67%
Coverage XML written to file coverage.xml

============================== 101 passed, 1 warning in 3.83s ==============================
_________________________________________ summary __________________________________________
  py38: commands succeeded
  congratulations :)
```

And this file definitely doesn't have a line number "3726" in `cli.py`:

```
pycobertura/cli.py                    109     81    26%   268-3726
```

As you can see above there is a deprecation warning about `xunit1` vs. `xunit2`. I ran a `pip freeze` between both versions of py37 and py38 but the versions of `coverage` and `pytest` seem to be the same in both pyenvs.

image

So that _seems_ unrelated. I wonder if Python 3.7 and 3.8 measure line hits/misses in a different way... ? Maybe we should try to upgrade `coverage` and `pytest-cov` and have pycobertura work on more recent versions... `¯\_(ツ)_/¯`

A while ago, we pinned versions of `coverage` and `pytest-cov` in `test-requirements.txt` because later versions were causing pycobertura to fail due to problematic data in the generated coverage file:

```
cat test-requirements.txt | grep cov
coverage>=4.0,<4.4
pytest-cov<2.6.0
```

I found the ticket that required me to pin to older versions here: https://github.com/nedbat/coveragepy/issues/578 (the original report was on bitbucket here: https://bitbucket.org/ned/coveragepy/issues/578/incomplete-file-path-in-xml-report)

nilleb:
You should be good to go.

I had no problem removing the version spec from coverage and pytest, at least not with py37 and py38, and this has brought the _missing lines_ to a valid interval.

I have added an option to avoid the pytest warning to the setup.cfg.

But I have reverted (dropped) a call to shlex, because it prevented the subprocess.call from functioning.

```
----------------------------------- generated xml file: /Users/ivo/dev/pycobertura/junit.xml ------------------------------------

---------- coverage: platform darwin, python 3.8.2-final-0 -----------
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
pycobertura/__init__.py                 2      0   100%
pycobertura/cli.py                     66      0   100%
pycobertura/cobertura.py              199      0   100%
pycobertura/filesystem.py              90      0   100%
pycobertura/reporters.py              164      0   100%
pycobertura/templates/__init__.py       0      0   100%
pycobertura/templates/filters.py        8      0   100%
pycobertura/utils.py                  111      1    99%   31
-----------------------------------------------------------------
TOTAL                                 640      1    99%
Coverage XML written to file coverage.xml
```

nilleb:
And, good luck for job hunt.

aconrad:
Thanks, @nilleb, for this wonderful PR. I don't think we're quite ready to unpin the test requirements after all but I wanted to get this off your plate so I can keep investigating on my side.

PR #97, opened 2020-05-29
Repin coverage and pytest cov to older versions

aconrad:
Reverting to using older versions of `pytest-cov` and `coverage.py` because of this issue: https://github.com/nedbat/coveragepy/issues/578

Although code coverage reports generated via py38 doesn't work well unless we use the latest versions of `pytest-cov` and `coverage.py`.

I don't really know what the XML Cobertura format is _supposed_ to be and what coverage.py might be different than other coverage tools out there that also generate coverage reports in the Cobertura formats.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
     - pip install tox-travis
 
 # Because the tests access the first git commit, we need to fetch the whole
-# repository without using a git --depth parameter. See
+# pycobertura repository without using a git --depth parameter. See
 # https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth
 git:
   depth: false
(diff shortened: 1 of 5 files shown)


nilleb:
The errors you get on the CI come from the fact that the source code is getting fetched with a max depth of 50
While on your repository you have about 200 commits.

```
git clone --depth=50 https://github.com/aconrad/pycobertura.git aconrad/pycobertura
```

You could consider this flag:
https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

aconrad:
Oh whoa, that's useful, thanks for the tip @nilleb! I've been trying to run a Travis build in debug mode because it wasn't obvious what was going on, but I've failed to get the debug mode to work despite going through the instructions here https://docs.travis-ci.com/user/running-build-in-debug-mode/

PR #99, opened 2020-06-30
Fix coverage diff reporting when only a single report is available

bastih:
While toying with a fallback case for the single-coverage file option, I found that the coverage was misrepresented for several lines because the coverage status for the source1 is read from the cobertura file that only has the lines for source2.

I added a testcase where I'm starting with some uncovered files and add some more uncovered code which results in incorrect coverage reports.

This issue occurs in a scenario where we run

pycobertura diff -source1 prevsrc/ --source2 currentsource/ coverage.xml coverage.xml

and coverage.xml was created on currentsource (as described in https://github.com/aconrad/pycobertura#i-only-have-one-cobertura-report-can-i-just-see-my-uncovered-changes)

In the results I noticed that the diff was showing coverage changes in lines that had not changed from source1 to source2 and that definitely hadn't experienced a change in coverage (these were lines in one long function that would always execute). While digging into the code, I noticed that lines do get matched up between the two sources, but lines are not adjusted in coverage statuses, which lead to incorrect changes showing up in this scenario.

Pycobertura version: 0.10.5

Coverage in our case is generated by gcovr from C++ code, but in the PR I reproduced the same behavior with python.


aconrad:
Thanks @bastih ! Your contribution is much appreciated. Can you update the PR description with more details on what the problem you are having is? Maybe sample pycobertura commands you ran?

Also, can you please provide what versions of coverage.py (the version that generated the coverage report) and Python you are using?

I will take a look eventually, but before I do I want to make sure you can provide as much context as possible about what you are encountering.

bastih:
Added some more details to PR.

tests/test_cobertura_diff.py
```diff
@@ -95,6 +95,17 @@ def test_diff_line_rate_by_class_file():
             expected_sources[filename]
 
 
+
```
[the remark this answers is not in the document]
bastih: Will do!

pycobertura/cobertura.py
```diff
@@ -368,6 +371,13 @@ def file_source(self, filename):
         # Build a dict of lineno2 -> lineno1
         lineno_map = reconcile_lines(lines2, lines1)
 
+        # when we are using the same coverage file for both sides, we need to
+        # translate the coverage of lines1 so that it corresponds to its
+        # real lines.
+        if self.cobertura1 == self.cobertura2:
+            rev_map = {v: k for k, v in lineno_map.items()}
+            line_statuses1 = {rev_map[v]: v for v in lineno_map.values()}
```
aconrad, line 379: I'm looking a little closer at this code, and wonder... It seems that `line_statuses1` doesn't hold the same value types as before. When I add `print()` statements like this:

```diff
diff --git a/pycobertura/cobertura.py b/pycobertura/cobertura.py
index 75315b7..24c26fb 100644
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -375,8 +375,13 @@ class CoberturaDiff(object):
         # translate the coverage of lines1 so that it corresponds to its
         # real lines.
         if self.cobertura1 == self.cobertura2:
+            print()
+            print('line_statuses1 (pre):', line_statuses1)
+            print('lineno_map:', lineno_map)
             rev_map = {v: k for k, v in lineno_map.items()}
+            print('rev_map:   ', rev_map)
             line_statuses1 = {rev_map[v]: v for v in lineno_map.values()}
+            print('line_statuses1 (post):', line_statuses1)
 
         lines = []
         for lineno, source in enumerate(lines2, start=1):
```

... then the output of the new test looks like this:

```
$ pytest -sk test_diff_same_report_different_source_dirs

...

tests/test_cobertura_diff.py::test_diff_same_report_different_source_dirs 
line_statuses1 (pre): {1: True, 2: True, 3: False, 4: True, 6: True, 7: True, 8: False, 9: True, 11: True, 12: True}
lineno_map: {1: 1, 4: 2, 5: 3, 6: 4, 7: 5, 8: 6, 9: 7, 10: 8, 11: 9, 12: 10}
rev_map:    {1: 1, 2: 4, 3: 5, 4: 6, 5: 7, 6: 8, 7: 9, 8: 10, 9: 11, 10: 12}
line_statuses1 (post): {1: 1, 4: 2, 5: 3, 6: 4, 7: 5, 8: 6, 9: 7, 10: 8, 11: 9, 12: 10}
```

Normally, `line_statuses1` should contain the line number as the key and a boolean as the value to show whether the line has been covered, but here we have line numbers as keys and values. Can you elaborate on your thought process here?

bastih: ugh, this must have gone missing in a refactor... let me fix that o_O

bastih: Ah this has gotten lost in local refactorings I guess. It's odd that this would not break other tests. Now checking if I can cover this better (along with the pre-refactoring version --- which I like more for readability anyway).

bastih:
Sorry for that - this is what I get for preemptively moving code around and not re-checking it.

aconrad:
@bastih, wonderful, thanks for making this work! It looks great! I'll go ahead and merge this! I'll publish it likely tomorrow if that works for you.

bastih:
Awesome! Thanks for working with me on this!

aconrad:
https://pypi.org/project/pycobertura/1.0.1/

Issue #102, opened 2020-07-25
Can I use it with preload xml content?


williamfzc:
Hi there, good project!

In my cases, I used some snippets to convert lcov xml to Cobertura XML. And it will generate xml-format string.

```python
converter = LcovCobertura(info_content)
cobertura_xml = converter.convert()

# now I have a xml-format string
```

So can we directly use xml-format string (rather than write it to file and read from it in pycobertura)?
Or if you prefer a PR? :)

aconrad:
I'm sure we can change that in some way to to prevent you from doing this intermediary step. It wasn't designed to support this use case but I'm sure we could accommodate without too much fuss. Have you looked at the code to see if there are opportunities for improvement?

williamfzc:
> Have you looked at the code to see if there are opportunities for improvement?

yep, I think it may be not that hard. (it looks like)

https://github.com/aconrad/pycobertura/blob/cdce1ae0eacd6d8b2954d289a2379330967abf97/pycobertura/cobertura.py#L40

xml content will be parsed at the beginning of `__init__`. We can add a pre-check here to identify if it is a valid xml-string. If it is not, treat it as a path.

aconrad:
Thanks for this @williamfzc! Can you please submit a PR so we follow the standard code reviewing process?

williamfzc:
> Thanks for this @williamfzc! Can you please submit a PR so we follow the standard code reviewing process?

Sure :) see #103

aconrad:
Addressed in #103, closing.

PR #103, opened 2020-08-06
feat: support xml parser from string

williamfzc:
#102

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Support loading Cobertura reports from an XML string. Thanks @williamfzc
+
 ## 1.0.1 (2020-07-08)
 
 * Fix misreported coverage when a single coverage file is used with
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -1,4 +1,6 @@
 import lxml.etree as ET
+import lxml
+import os
 
 from collections import namedtuple
 
@@ -47,7 +49,15 @@ class Cobertura(object):
         files referenced in the report. Please check the pycobertura.filesystem
         module in order to discover more about filesystems.
         """
-        self.xml = ET.parse(report).getroot()
+        try:
+            self.xml = ET.fromstring(report).getroot()
+        except lxml.etree.XMLSyntaxError:
+            # it's not a xml-string, so it should be a path.
+            # is it existed?
+            if not os.path.isfile(report):
+                raise FileNotFoundError("file not existed: {}".format(report))
+            # it is an existed file.
+            self.xml = ET.parse(report).getroot()
         self.report = report if isinstance(report, basestring) else None
 
         if filesystem:
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -1,11 +1,9 @@
 import lxml.etree as ET
-import lxml
 import os
 
 from collections import namedtuple
 
 from pycobertura.filesystem import filesystem_factory
-
 
 from pycobertura.utils import (
     extrapolate_coverage,
@@ -39,6 +37,9 @@ class Cobertura(object):
     An XML Cobertura parser.
     """
 
+    class InvalidCoverageReport(Exception):
+        pass
+
     def __init__(self, report, filesystem=None):
         """
         Initialize a Cobertura report given a coverage report `report`. It can
@@ -50,14 +51,10 @@ class Cobertura(object):
         module in order to discover more about filesystems.
         """
         try:
-            self.xml = ET.fromstring(report).getroot()
-        except lxml.etree.XMLSyntaxError:
-            # it's not a xml-string, so it should be a path.
-            # is it existed?
-            if not os.path.isfile(report):
-                raise FileNotFoundError("file not existed: {}".format(report))
-            # it is an existed file.
-            self.xml = ET.parse(report).getroot()
+            self.xml = self._load_from_string(report)
+        except ET.XMLSyntaxError:
+            self.xml = self._load_from_path(report)
+
         self.report = report if isinstance(report, basestring) else None
 
         if filesystem:
@@ -67,6 +64,14 @@ class Cobertura(object):
 
     def __eq__(self, other):
         return self.report and other.report and self.report == other.report
+
+    def _load_from_path(self, report_path):
+        if not os.path.isfile(report_path):
+            raise self.InvalidCoverageReport("Invalid coverage file: {}".format(report_path))
+        return ET.parse(report_path).getroot()
+
+    def _load_from_string(self, s):
+        return ET.fromstring(s)
 
     @memoize
     def _get_class_element_by_filename(self, filename):
--- a/tests/test_cobertura.py
+++ b/tests/test_cobertura.py
@@ -1,5 +1,6 @@
 import mock
 import pytest
+import lxml.etree as ET
 
 from .utils import make_cobertura
 
@@ -7,7 +8,7 @@ from .utils import make_cobertura
 def test_parse_path():
     from pycobertura import Cobertura
 
-    xml_path = 'foo.xml'
+    xml_path = 'tests/cobertura.xml'
 
     with mock.patch('pycobertura.cobertura.ET.parse') as mock_parse:
         cobertura = Cobertura(xml_path)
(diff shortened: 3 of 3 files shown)


williamfzc:
I would fix these cases in some hours. Maybe tomorrow.
And then you can start reviewing it.

pycobertura/cobertura.py
```diff
@@ -47,7 +49,15 @@ def __init__(self, report, filesystem=None):
...
-        self.xml = ET.parse(report).getroot()
+        try:
+            self.xml = ET.fromstring(report).getroot()
+        except lxml.etree.XMLSyntaxError:
+            # it's not a xml-string, so it should be a path.
+            # is it existed?
+            if not os.path.isfile(report):
+                raise FileNotFoundError("file not existed: {}".format(report))
+            # it is an existed file.
+            self.xml = ET.parse(report).getroot()
```
aconrad, line 60: The method `__init__()` is getting a little crowded... Can we break out the logic into methods such as `_load_from_string` and `_load_from_path`?

```suggestion
        try:
            self.xml = self._load_from_string(report)
        except lxml.etree.XMLSyntaxError:
            self.xml = self._load_from_path(report)
```

... and then abstract away the logic in the respective methods. What do you think?

pycobertura/cobertura.py
```diff
@@ -47,7 +49,15 @@ def __init__(self, report, filesystem=None):
...
         module in order to discover more about filesystems.
         """
-        self.xml = ET.parse(report).getroot()
+        try:
+            self.xml = ET.fromstring(report).getroot()
+        except lxml.etree.XMLSyntaxError:
+            # it's not a xml-string, so it should be a path.
+            # is it existed?
+            if not os.path.isfile(report):
+                raise FileNotFoundError("file not existed: {}".format(report))
```
aconrad, line 58: Can we raise a specific exception? Maybe `InvalidCoverageReport`? We could add this exception under the `Cobertura` class namespace.

```python
class Cobertura(object):
    class InvalidCoverageReport(Exception):
        pass
    ...
```

And then use that to Can we tweak the wording here?

```suggestion
                raise self.InvalidCoverageReport("Invalid coverage report: {}".format(report))
```

pycobertura/cobertura.py
```diff
@@ -47,7 +49,15 @@ def __init__(self, report, filesystem=None):
         files referenced in the report. Please check the pycobertura.filesystem
         module in order to discover more about filesystems.
         """
-        self.xml = ET.parse(report).getroot()
+        try:
+            self.xml = ET.fromstring(report).getroot()
+        except lxml.etree.XMLSyntaxError:
```
aconrad, line 54: Can we reuse the module `ET`?

```suggestion
        except ET.XMLSyntaxError:
```

Or stop using `ET` and use `lxml.etree` instead? I don't have a strong preference, but let's use one or the other for consistency.

pycobertura/cobertura.py
williamfzc: Approve. I prefer making them a closure because they will only be used in this method:

```python
def _load_from_string() ...
def _load_from_path() ...

try:
    self.xml = _load_from_string(report)
except lxml.etree.XMLSyntaxError:
    self.xml = _load_from_path(report)
```

Both are ok I think.

pycobertura/cobertura.py
williamfzc: Ok

pycobertura/cobertura.py
williamfzc: oh sorry about that. I have forgotten `ET`.
I would replace `lxml.etree` with `ET`.

pycobertura/cobertura.py
williamfzc: I noticed that you have added a specific exception in FileSystem.
Maybe we can add a `exceptions` module to manage them like this? 
https://github.com/psf/requests/blob/master/requests/exceptions.py

It looks clear.

pycobertura/cobertura.py
aconrad: Hm. I was trying to visually declutter `__init__()`, so making them closures would defeat my initial intent.

In Python, I tend to use closures to prevent me from writing small classes, to carry state when I'm only using functions. But here we're already in the context of a full-fledged class so would adding closures add much value? If the purpose is because they aren't used anywhere else, I actually like to create functions/methods that aren't necessarily shared. I think they bring additional readability to the code, thanks to their function's name. I'd rather reuse the existing mechanics of the class instead of introducing a pattern of closures in this code.

What do you think?

williamfzc:
I have done as comment said. It looks good ~

Another small question:
where is `foo.xml`? Test cases will pass if I change it to `tests/cobertura.xml`.
https://github.com/aconrad/pycobertura/blob/cdce1ae0eacd6d8b2954d289a2379330967abf97/tests/test_cobertura.py#L10

pycobertura/cobertura.py
```diff
@@ -47,7 +52,11 @@ def __init__(self, report, filesystem=None):
         files referenced in the report. Please check the pycobertura.filesystem
         module in order to discover more about filesystems.
         """
-        self.xml = ET.parse(report).getroot()
+        try:
+            self.xml = self._load_from_string(report)
+        except ET.XMLSyntaxError:
+            self.xml = self._load_from_path(report)
```
aconrad, line 58: Looking good! 🎉

pycobertura/cobertura.py
```diff
@@ -58,6 +67,17 @@ def __init__(self, report, filesystem=None):
     def __eq__(self, other):
         return self.report and other.report and self.report == other.report
 
+    def _load_from_path(self, report_path):
+        # it's not a xml-string, so it should be a path.
+        # is it existed?
+        if not os.path.isfile(report_path):
+            raise self.InvalidCoverageReport("file not existed: {}".format(report_path))
```
aconrad, line 74: Can we tweak the wording?

```suggestion
            raise self.InvalidCoverageReport("Invalid coverage file: {}".format(report_path))
```

pycobertura/cobertura.py
```diff
@@ -58,6 +67,17 @@ def __init__(self, report, filesystem=None):
...
+        # it is an existed file.
```
aconrad, line 75: Is the comment necessary? I think the code is self-explanatory enough.

```suggestion
```

tests/test_cobertura.py
```diff
@@ -15,6 +16,15 @@ def test_parse_path():
     assert cobertura.xml is mock_parse.return_value.getroot.return_value
 
 
+def test_parse_string():
+    from pycobertura import Cobertura
+
+    xml_path = 'foo.xml'
+    with open(xml_path) as f:
+        xml_string = f.read()
+    assert ET.tostring(Cobertura(xml_path).xml) == ET.tostring(Cobertura(xml_string).xml)
```
aconrad, line 25: Good test! 👍

pycobertura/cobertura.py
```diff
@@ -58,6 +67,17 @@ def __init__(self, report, filesystem=None):
     def __eq__(self, other):
         return self.report and other.report and self.report == other.report
 
+    def _load_from_path(self, report_path):
+        # it's not a xml-string, so it should be a path.
+        # is it existed?
```
aconrad, line 72: Is this comment relevant here? Maybe we can remove it, the code is pretty self-explanatory to me.

```suggestion
```

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Support loading xml from string directly. Thanks @williamfzc
```
aconrad, line 5:
```suggestion
* Support loading Cobertura reports from an XML string. Thanks @williamfzc
```

aconrad:
> I have done as comment said. It looks good ~

Thank you!!

> Another small question:
> where is `foo.xml`? Test cases will pass if I change it to `tests/cobertura.xml`.

The file `foo.xml` doesn't exist. This test doesn't really care about the presence of an XML file, it's mocking the parser and tests that Cobertura calls the XML parser's `.getroot()` method. I'd argue that this test is testing the internals of how Cobertura parses the file, so I wouldn't say it's a good practice. I should probably clean it up.

williamfzc:
Ok I have finished it. Thanks for your review :)

> The file foo.xml doesn't exist. This test doesn't really care about the presence of an XML file, it's mocking the parser and tests that Cobertura calls the XML parser's .getroot() method. I'd argue that this test is testing the internals of how Cobertura parses the file, so I wouldn't say it's a good practice. I should probably clean it up.

Finally I change `foo.xml` to `tests/cobertura.xml` for making all cases passed.

aconrad:
Wonderful, @williamfzc! Thank you so much for all your work on this! 🙏

I realize we forgot to update the docstring of `__init__()`, but I'll merge your PR as-is a take care of it myself, don't worry about it. 👍

Issue #108, opened 2020-12-02
Per directory stats?


dothebart:
I see that `show` can print out per file summary. I'd need this per directory - is there already a way to do this?

aconrad:
Hi @dothebart, pycobertura treats all file paths as strings, so it's currently agnostic of what a file vs. a directory is.

Are you using pycobertura from the command line or as a library?

dothebart:
I've did first experiments with the CLI. 
I'm thinking about copying the `class TextReporter(Reporter):` , keep the stats on the object, and print them out once the path component of `filename` changes?

aconrad:
Wondering if `Cobertura.files()` could take a keyword argument to allow it to list only a subset of files from the report that would match a given pattern. 🤔

```python
Cobertura.files(match="*.py")

# or

Cobertura.files(match="foo/bar/*")
```

Dunno, food for thought. 😃

PR #109, opened 2020-12-03
Feature/per directory coverage

dothebart:
as discussed in https://github.com/aconrad/pycobertura/issues/108 I need another formatter that groups sources by directory and other criteria. 

I would make this a draft pr since this is still very hacky, but for some reason github won't let me. 

the yml for grouping looks like this:
```
Cache:
  coverage/arangod/Cache
Cluster:
  coverage/arangod/Cluster
  coverage/arangod/Network
  coverage/arangod/Sharding
```

### DIFF
--- a/pycobertura/cli.py
+++ b/pycobertura/cli.py
@@ -48,7 +48,7 @@ def get_exit_code(differ, source):
 
 @pycobertura.command()
 @click.argument("cobertura_file")
-@click.option("-f", "--format", default="textdir", type=click.Choice(list(reporters)))
+@click.option("-f", "--format", default="text", type=click.Choice(list(reporters)))
 @click.option(
     "-o",
     "--output",
--- a/pycobertura/cli.py
+++ b/pycobertura/cli.py
@@ -4,6 +4,7 @@ from pycobertura.cobertura import Cobertura
 from pycobertura.reporters import (
     HtmlReporter,
     TextReporter,
+    TextDirectorySummaryReporter,
     HtmlReporterDelta,
     TextReporterDelta,
 )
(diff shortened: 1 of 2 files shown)


aconrad:
Hi @dothebart, thanks for your pull request!

I took a quick glance at your changes and I think I better understand what you are trying to do; although having a screenshot of the output in the PR description would definitely help. I also acknowledge that you said that it was still a draft.

From my point of view, I worry that this new reporter is a little too specific to your use case and wonder if it should be part of pycobertura. I tried to design pycobertura in a way that allows extensibility by creating custom reporter like you did. I don't have a lot of experience extending the CLI portion of pycobertura, but wouldn't it be possible for you to keep the implementation of a new reporter in its own repository? Say `pycobertura-directory-reporter-plugin` and then use that to extend the behavior of pycobertura?

I would be happy to take changes to pycobertura to accommodate its extensibility, however. Again, without much experience extending pycobertura myself, there's likely room to make it easier on users. For example, I could see us introduce a way to register new formatters for the command line. Formatters are currently registered in a dictionary `{"<reporter_name>": <reporter class>}` (like you did in this PR) and there might be better ways to modify that object without monkey patching a global variable.

What do you think?

dothebart:
Added the sample output. 

In general you see that it stort of plugs nicely into the reporters interface, so having a CLI-Reporter plugin would probably exactly fit my needs. 
The code here still all a bit rough around the corners, since I was a bit under time pressure, and working with the [20Megs xml](https://github.com/arangodb/arangodb) took several minutes on each try ;-) So I didn't add tests or such yet.

So I thought we can use this as a discussion start rather than theoretically talking about what could or should be. 
I found nice about the CLI that I didn't have to do much to get my things done, it mostly doesn't even interfere 
with the rest of the code aside of deriving a readily existing base class.

aconrad:
Absolutely! Totally open for discussion. I'm sharing preliminary thoughts about what crossed my mind at a first glance.

And if you see performance improvements along the way since you are dealing with a large XML file, we welcome change here, too! 😉

PR #113, opened 2021-03-12
handle multiple classes in the same file, Fixes #112 

### LINKED ISSUE #112: Multiple classes in the same file are ignored
It looks like if a cobertura report has multiple classes in the same file, only the first class will populate the stats for that file and the others will be ignored. 
[coverageExample.zip](https://github.com/aconrad/pycobertura/files/6131849/coverageExample.zip)

smortvedt:
I tried to add this feature while minimally changing the other features. I very new to this project so please review carefully as I just tested a few different cases. Fixes #112

### DIFF
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -78,14 +78,26 @@ class Cobertura(object):
         return ET.fromstring(s)
 
     @memoize
+    def _get_class_elements_by_filename(self, filename):
+        syntax = "./packages//class[@filename='%s']" % (filename)
+        classes = []
+        for c in self.xml.xpath(syntax):
+            classes.append(c)
+        return classes
+        
+    @memoize
     def _get_class_element_by_filename(self, filename):
         syntax = "./packages//class[@filename='%s'][1]" % (filename)
         return self.xml.xpath(syntax)[0]
 
     @memoize
     def _get_lines_by_filename(self, filename):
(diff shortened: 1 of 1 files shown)


gro1m:
Hi @aconrad @smortvedt 
This is a mistake that we should fix.

Could you (@smortvedt)  fix this maybe using the newest Python 3.7 possibilities, i.e. f-strings et cetera in cobertura.py. In addition, one has to change the testing checks in cli.py that fail according to the TRAVIS CI output:
https://app.travis-ci.com/github/aconrad/pycobertura/jobs/490695222#L400-L484.

The mistake is apparent when looking at `tests/cobertura.xml`: one sees that the statistics from lines 52-70 from the second class with ""Main$Helper" are not included, i.e. lines with numbers from 31-34 and thus there should also be 15 instead of 11 statements...

I could also do this, but @smortvedt would need to give me developer rights on his master branch if I were to do it in this PR. Otherwise, I'd need to pull the branch separately and integrate it into my fork and work from there...

aconrad:
This totally fell through the cracks, I'm sorry about that. I'll try to take a look at this proposal this week.

aconrad:
@smortvedt any chance we can merge the latest code from master into your branch?

aconrad:
Closing this PR in favor of #132 where it's being revived. Thanks @smortvedt for your work here!

PR #123, opened 2022-01-08
Maintenance/update to python37

gro1m:
This PR updates the project to currently supported Python versions, i.e.:
Python 3.7, 3.8, 3.9 and 3.10.
It also includes the complete migration from `setup.py` to the static `setup.cfg` and `pyproject.toml` to keep-up with current standards. The `setup.py` shall only contain dynamic content, which was introduced by using pypandoc previously. This package is not needed anymore, as PyPI now supports markdown natively.


release.sh
```diff
@@ -1,15 +1,13 @@
 # Before you release
-# - install pandoc on your system: apt-get install pandoc
-# - install pypandoc: pip install pypandoc
 # - install twine: pip install twine
-# - bump package version in setup.py
+# - bump package version in setup.cfg
 # - update the package version in the CHANGES file
 # - commit the changes to master and push
 
-PKG_NAME=$(python setup.py --name)
```
aconrad, line 9: Is the output of this command just `pycobertura`? If that's the case, we can probably just hardcode it instead of having a code trying to extract it. What do you think?

release.sh
```diff
@@ -1,15 +1,13 @@
...
-PKG_VERSION=$(python setup.py --version)
+PKG_NAME=$(python read_setup_cfg.py --section metadata --key name)
+PKG_VERSION=$(python read_setup_cfg.py --section metadata --key version)
```
aconrad, line 8: It's kind of a bummer we have to write `read_setup_cfg.py`... I was browsing around and found this to extract the version:

```python
>>> from importlib import metadata
>>> metadata.version('pycobertura')
'2.1.0'
```

But I don't know if that would be available by the time we try to run this code (chicken-and-egg problem?)

aconrad: I also stumbled upon this about [single-sourcing the package version](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/). Would it be relevant to what we're trying to achieve?

gro1m: I think actually even better would be the following:
```
from setuptools.config import read_configuration

conf_dict = read_configuration("/home/user/dev/package/setup.cfg")
```
Reference: https://setuptools.pypa.io/en/latest/setuptools.html.

I will test this locally and check in again.
Also with regards to the Python 3.7 problems we have (maybe it is the pip version, maybe it it is the python version that still needs a setup.py in editable mode).

gro1m: Unfortunately only supported from Python 3.8 onwards, but we can use `pkg_resources`. Python 3.7 is still supported for another 1 and a half years, so we should probably support it. This is also why we still need a empty setup.py with just an empty setup function call - although locally it works without

gro1m:
Hi @aconrad 
So I think in both PRs the Python3.10 tests do not work and I think it is due to the build configuration, as Ubuntu 16.04 Xenial is deprecated. So we would need minimally Ubuntu 18.04 or Ubuntu 20.04 (LTS versions would of course be better to have, i.e. Bionic Beaver or Focal Fossa). Can you do something there?

aconrad:
@gro1m can we remove 3.10 support? I feel Travis isn't ready for it, I don't see it listed here:

https://docs.travis-ci.com/user/languages/python/

And people are talking about here, too: https://travis-ci.community/t/add-python-3-10/12220/11

.gitignore
```diff
@@ -57,3 +57,6 @@ docs/_build/
 # Editors
 *.swp
 .vscode
+
+# Mac OS
+.DS_Store
```
aconrad, line 62: Nit, but can you add a new line to the end of files? I have my VS Code configured to always add a new line at the end of files upon saving so I don't have to think about it.

Screen Shot 2022-01-12 at 11 08 49 AM

.travis.yml
```diff
@@ -1,12 +1,14 @@
...
-    - python: 3.6
-      env: TOXENV=py36
     - python: 3.7
       env: TOXENV=py37
+    - python: 3.8
+      env: TOXENV=py38
+    - python: 3.9
+      env: TOXENV=py39
+    - python: 3.10
+      env: TOXENV=py310
```
aconrad, line 11: Let's drop support for 3.10 until Travis is ready.

release.sh
```diff
@@ -1,15 +1,14 @@
 # Before you release
-# - install pandoc on your system: apt-get install pandoc
-# - install pypandoc: pip install pypandoc
 # - install twine: pip install twine
-# - bump package version in setup.py
+# - bump package version in setup.cfg
```
aconrad, line 3: This for adapting the comments 🙏

release.sh
```diff
@@ -1,15 +1,14 @@
...
 # - update the package version in the CHANGES file
 # - commit the changes to master and push
 
-PKG_NAME=$(python setup.py --name)
-PKG_VERSION=$(python setup.py --version)
+PKG_NAME = pycobertura
```
aconrad, line 7: I dont think we can have spaces around the `=` sign:

```zsh
$ PKG_NAME = pycobertura
zsh: command not found: PKG_NAME
```

But this works:

```zsh
$ PKG_NAME=pycobertura 
$ echo $PKG_NAME 
pycobertura
```

setup.cfg
```diff
@@ -29,3 +32,49 @@ addopts =
...
+    License :: OSI Approved :: MIT License
+    Natural Language :: English
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
aconrad, line 60: Can we remove 3.10 until we have a working Travis build?

README.md
```diff
@@ -36,8 +36,8 @@ NOTE: The API is unstable any may be subject to changes until it reaches 1.0.
 
 ## Install
 
-```
-$ pip install pycobertura
+```bash
+pip install pycobertura
```
aconrad, line 40: Should this command be `pip3 install pycobertura`?

release.sh
```diff
@@ -1,15 +1,14 @@
...
+# from importlib import metadata is unfortunately only available from Python 3.8 onwards
+PKG_VERSION = python -c "from pkg_resources import get_distribution; print(get_distribution('pycobertura').version)"
 
-git tag -am "release v${PKG_VERSION}" v${PKG_VERSION}
+git tag -am "release v${PKG_VERSION}" "v${PKG_VERSION}"
 git push --tags
-python setup.py sdist
-twine upload dist/${PKG_NAME}-${PKG_VERSION}.tar.gz
+python -m build
+twine upload dist/"${PKG_NAME}-${PKG_VERSION}*"
```
aconrad, line 14: What is the reason for changing this to a wildcard instead of the more explicit `.tar.gz`?

aconrad (commented):
This is looking great! Thanks!

I left a few comments, can you please take a look?

release.sh
gro1m: yes you are right the shell does not allow whitespaces around = (I am so used to this in the other programming languages that I forget, that the shell does not allow this)

setup.cfg
gro1m: yes

.travis.yml
gro1m: ok

.gitignore
gro1m: thanks for the tip

.travis.yml
gro1m: ok

setup.cfg
gro1m: yes

release.sh
gro1m: I think we should also upload the .whl file, see the following article: https://realpython.com/python-wheels/
Principally the documentation even recommends `twine upload dist/*`. 
But we could also just do the .tar.gz and the .whl in 2 lines.

README.md
gro1m: I do not think that it is really necessary, especially as if you only have Python 3 installed the commands `python` and `pip` should work without any configuration. Also normally people work with virtual environments these commands also will always work as they should. But if you want to make clear that Pycobertura only supports Python3 we could also change it.

release.sh
aconrad: Oh, interesting... Yeah, then let's wheel it up!

Can we be more explicit about the `.whl` format? Maybe something like `twine upload dist/${PKG_NAME}-${PKG_VERSION}.{tar.gz,whl}`. The only reason to make it more explicit is to remember to consider the wheel file next time we need to modify this script.

aconrad (commented):
Looks good! I left one last comment but we're close to merging this change! 😍

aconrad:
Oh, and last but not least: can you update the CHANGES that lists the changes done under the "Unreleased" paragraph? And mention yourself, see past contributions as a template. 😉

gro1m:
Hi @aconrad 
I hope it looks good now. There is one last thing: You merged the CHANGES.md and the README.md, which I do not do. Doing this also leads to the "problem" that you would use twice the "#"-heading (the title). So I am not quite sure what you want to do there. I am not really sure if there is an easy option to merge those in the `setup.cfg` and I am not sure what the correct convention is regarding the title heading. From my viewpoint, it would be enough to have a link to the release notes, but I am not quite sure what you want to do there :)

gro1m:
> > There is one last thing: You merged the `CHANGES.md` and the `README.md`, which I do not do anymore. [...] I do not know yet of an easy option to merge those in the `setup.cfg` [...]
> 
> Can we do this to concatenate the files?
> 
> ```
> long_description = file: README.md, CHANGES.md
> ```
> 
> This seems to be a valid syntax according to: https://setuptools.pypa.io/en/latest/userguide/declarative_config.html
> 
> What do you think? Or is there something you said I missed? Not sure I understand the part with headings.

That's it - I remembered something, but then I did not find it anymore and I was a bit sceptical when I read that the argument had to be of type string (I did not sink of a comma-delimited string).

The thing with the headings is that you should actually not use heading of level 1 ("title") in a markdown document twice, at least a markdown linter will complain, see also: https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md025.
But I can live with that.

aconrad:
Thanks for everything you've done here, i very much appreciate it.

PR #124, opened 2022-01-08
Refactor/reporters

gro1m:
This PR is mainly a refactoring of the `reporters.py` file. 
It shifts as much configuration and formatting into the base classes `Reporter` and `DeltaReporter`. The main reason for this is to keep the needed changes to this file when introducing new formats, which needs a Reporter as well as a DeltaReporter subclass, minimal! Furthermore, it sets a standard for all the output tables.
 Possible future supportable output formats include: JSON, YAML, CSV, TSV, NSV (new-line separated values sometimes also summarized as TSV).

Also, the refactoring aaims to show clearly the used data structure, which are named tuples at the moment and follow a row-based logic. Namedtuples are immutable, as the code contains options to include a `missed_lines` column, the analogue mutable data structure `recordtype` might be an option.
Dictionaries could be an option as well (especially when trying to extend to JSON and YAML formats); the nature of dictionaries demand though to have unique keys and thus the code would have to be transformed in such a way as that it adheres to a column-wise implementation logic and then the output could be formatted as in the dictionary example from [tabulate](https://pypi.org/project/tabulate/) to have the headers as keys:
```python
>>> print(tabulate({"Name": ["Alice", "Bob"],
...                 "Age": [24, 19]}, headers="keys"))
  Age  Name
-----  ------
   24  Alice
   19  Bob
```


aconrad:
Can we rebase and resolve conflicts?

pycobertura/reporters.py
```diff
@@ -14,77 +14,54 @@
...
+        lines = {
+            "Filename": [filename for filename in self.cobertura.files()],
+            "Stmts": [
+                self.cobertura.total_statements(filename)
+                for filename in self.cobertura.files()
+            ],
+            "Miss": [
+                self.cobertura.total_misses(filename)
+                for filename in self.cobertura.files()
+            ],
```
aconrad, line 39: I'm trying to understand the performance implications of iterating over all filenames as many times as we have arrays to fill. Should we be worried?

I think ultimately we are doing the same amount of operations, but I'm wondering what the overhead cost is of iterating over those file names is – whereas before we did it in one pass.

pycobertura/reporters.py
```diff
@@ -93,152 +70,147 @@ def __init__(self, *args, **kwargs):
...
+                    ),
+                    self.format_total_misses(self.differ.diff_total_misses(filename)),
+                    self.format_line_rate(self.differ.diff_line_rate(filename)),
+                )
+                for filename in self.differ.files()
+                if any(
+                    (
+                        self.differ.diff_total_statements(filename),
+                        self.differ.diff_total_misses(filename),
+                        self.differ.diff_line_rate(filename),
```
aconrad, line 138: These are already calculated in the tuple comprehension above. Is there a way we compute them once and reuse them? It might require us to break down this "one-liner" comprehension.

pycobertura/reporters.py
```diff
@@ -249,61 +221,23 @@ def __init__(self, *args, **kwargs):
...
             "show_missing": self.show_missing,
             "show_source": self.show_source,
         }
 
         if self.show_source is True:
-            render_kwargs["sources"] = sources
+            render_kwargs["sources"] = [
+                (filename, self.differ.file_source_hunks(filename))
+                for filename in self.differ.files()
+                if self.differ.file_source_hunks(filename)
```
aconrad, line 240: This line is being computed twice, how can we avoid doing this?

pycobertura/reporters.py
```diff
@@ -14,77 +14,54 @@
 file_row = namedtuple("FileRow", row_attributes)
 file_row_missed = namedtuple("FileRowMissed", row_attributes + ["missed_lines"])
 
+headers_without_missing = ["Filename", "Stmts", "Miss", "Cover"]
+headers_missing = ["Filename", "Stmts", "Miss", "Cover", "Missing"]
```
aconrad, line 18: To be more symmetric with its counterpart `headers_without_missing`, would it make sense to rename `headers_missing` as such?

```suggestion
headers_with_missing = ["Filename", "Stmts", "Miss", "Cover", "Missing"]
```

aconrad (commented):
There's a lot going on in this PR. But my understanding is that we're changing the internal representation of how we pass report lines around. Instead of having an array of tuples (1 tuple per row), we now return a dict of arrays (1 key per column).

The library `tabulate` seems to support both ways column-wise and row-wise. Can you clarify the intent of this change again?

gro1m:
Hi @aconrad 
This PR is a bit more complicated. 
Up to commit https://github.com/aconrad/pycobertura/pull/124/commits/bb1aa1a60e6a90e7a9592fba43fbd97cfca444e5 it is a simple refactoring.
The following commits also make adjustments to the data structure, i.e. from named tuples to dictionaries.
The problem is at the moment that I only could do the reporter classes but not the delta reporters.
I have an implementation for the delta reporters but the test fails, because I have a hard time with the following lines:
- https://github.com/aconrad/pycobertura/blob/master/pycobertura/templates/html-delta.jinja2#L41
- https://github.com/aconrad/pycobertura/blob/master/pycobertura/templates/html-delta.jinja2#L46
- https://github.com/aconrad/pycobertura/blob/master/pycobertura/templates/html-delta.jinja2#L57

pycobertura/reporters.py
gro1m: Done

pycobertura/reporters.py
gro1m: List comprehensions should be fast, but what I did is store `self.cobertura.files()` as Reporter class variable.

pycobertura/utils.py
```diff
@@ -54,7 +54,7 @@ def __call__(self, *args, **kw):
 
 def colorize(text, color):
     color_code = ANSI_ESCAPE_CODES[color]
-    return "%s%s%s" % (color_code, text, ANSI_ESCAPE_CODES["reset"])
+    return f'{color_code}{text}{ANSI_ESCAPE_CODES["reset"]}'
```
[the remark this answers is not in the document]
gro1m: Yes, I think we do :)

pycobertura/reporters.py
gro1m: done

pycobertura/reporters.py
```diff
@@ -14,77 +14,54 @@
...
-        footer = file_row_missed(
-            "TOTAL",
-            self.cobertura.total_statements(),
-            self.cobertura.total_misses(),
-            self.cobertura.line_rate(),
-            [],  # dummy missed lines
-        )
-        lines.append(footer)
+        lines = {
+            "Filename": [filename for filename in self.cobertura.files()],
```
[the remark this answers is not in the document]
gro1m: we can't but I do not know why -> some tests will fail...

pycobertura/reporters.py
gro1m: maybe...

pycobertura/reporters.py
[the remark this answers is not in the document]
gro1m: but this irritates me :)

pycobertura/reporters.py
gro1m: I could do a list comprehension for each and then loop through the 3 lists to evaluate the `filenames_of_files_with_changes` - I am just not sure we gain much, but I can try to do if you want me to

gro1m:
Hi @aconrad 
I think now it is at a nice place (everything with dictionaries) and I resolved your review, but for the 2 non-resolved ones, which I commented above and are still open and would be good if you could go over the code again :)

aconrad:
Hey @gro1m, thanks again for your contributions. I'm away from a laptop for a couple of days, will take a look at your work next week.

pycobertura/reporters.py
gro1m: addressed in commit 503ec5e

pycobertura/reporters.py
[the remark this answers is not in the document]
gro1m: resolved in commit 6336de6

gro1m:
Hi @aconrad 
Thanks for the info, I should have now resolved addressed all of your comments above.

As an additional feedback: it would be nice if we could make the html report tests formatting-independent and if we could have a formatter for jinja2 files.
The lines 37-51 in `html-delta.jinja2` are quite confusing without proper indentation:
```html
{%- for iter in range(lines["Filename"]|length) %}
          <tr>
            <td><a href="#{{ lines["Filename"][iter] }}">{{ lines["Filename"][iter]}}</a></td>
            <td>{{ lines["Stmts"][iter] }}</td>
            <td><span{% if lines["Miss"][iter] | is_not_equal_to_dash %} class="{{ lines["Miss"][iter] | misses_color }}"{% endif %}>{{ lines["Miss"][iter] }}</span></td>
            <td>{{ lines["Cover"][iter] }}</td>
{%- if show_source and show_missing %}
            <td>
{%- for missed_line in lines["Missing"][iter] %}
            {%- if not loop.first %}, {% endif %}<span class="{{ missed_line | misses_color }}">{{ missed_line }}</span>
{%- endfor %}
            </td>
{%- endif %}
          </tr>
{%- endfor %}
```
It took a long time for me to see that the missed_line loop is nested inside the first for loop and that the if statement is in the first for loop...

aconrad:
> It took a long time for me to see that the missed_line loop is nested inside the first for loop and that the if statement is in the first for loop...

Ugh, yes. If I recall, I was trying to be strict about the string output. The dashes you see in `{%-` are meant to suppress the whitespace before the jinja template. It's probably not the best way to test it and it's prone to breakage / fragile. Best would probably be to parse the HTML output in an in-memory DOM representation and then check the DOM itself or dump the DOM back to a string for comparison with a fixture.

pycobertura/reporters.py
```diff
@@ -1,90 +1,64 @@
...
-file_row = namedtuple("FileRow", row_attributes)
-file_row_missed = namedtuple("FileRowMissed", row_attributes + ["missed_lines"])
+headers_without_missing = ["Filename", "Stmts", "Miss", "Cover"]
+headers_with_missing = ["Filename", "Stmts", "Miss", "Cover", "Missing"]
 
 
 class Reporter(object):
     def __init__(self, cobertura):
         self.cobertura = cobertura
+        self.files_info = self.cobertura.files()
```
aconrad, line 21: Just so you know, `self.cobertura.files()` is already cached/memoized, so calling it multiple times shouldn't incur additional XML parsing. Saying that in case you were using `self.files_info` as a cache.

pycobertura/reporters.py
```diff
@@ -93,152 +67,166 @@ def __init__(self, *args, **kwargs):
...
+            ],
+        }
 
-        if self.color is True and total_misses != "-":
-            colorize = [green, red][total_misses[0] == "+"]
-            total_misses = colorize(total_misses)
+        lines["Filename"].extend(["TOTAL"])
+        lines["Stmts"].extend(
+            [self.format_total_statements(self.differ.diff_total_statements())]
+        )
```
aconrad, line 187: Non-blocking: any reason for using `.extend()` in place of `+=`?

```suggestion
        lines["Stmts"] += [self.format_total_statements(self.differ.diff_total_statements())]
```

pycobertura/reporters.py
```diff
@@ -93,152 +67,166 @@ def __init__(self, *args, **kwargs):
...
+        if self.show_source:
+            diff_total_missing = [
+                self.differ.diff_missed_lines(filename)
+                for filename in self.differ_files_info
             ]
+            lines["Missing"] = [
+                self.format_missed_lines(diff_total_missing[i])
+                for i in indexes_of_files_with_changes
+            ]
+            lines["Missing"].extend([""])
```
aconrad, line 202: This would be a little cheaper since we don't have to construct a temporary array that would only hold the empty string:

```suggestion
            lines["Missing"].append("")
```

pycobertura/templates/html-delta.jinja2
```diff
@@ -34,16 +34,16 @@ table.code td.lineno pre {margin-right: 1rem;}
           </tr>
         </thead>
         <tbody>
-{%- for line in lines %}
+{%- for iter in range(lines["Filename"]|length) %}
```
aconrad, line 37: Is `iter` the file index in the array? Can we rename it to `file_index` or `file_idx`?

```suggestion
{%- for file_idx in range(lines["Filename"]|length) %}
```

What do you think?

pycobertura/templates/filters.py
```diff
@@ -17,6 +17,16 @@
 }
 
 
+def is_not_equal_to_dash(arg):
+    # print(f"is_not_equal_to_dash: arg={arg}")
```
aconrad, line 21: Can we clean up those comments?

```suggestion
```

tests/test_stringify.py
```diff
@@ -0,0 +1,52 @@
+def test_stringify_func__empty_bracket():
```
aconrad, line 1: This for writing tests for this new function! 👍

pycobertura/reporters.py
```diff
@@ -93,152 +67,166 @@ def __init__(self, *args, **kwargs):
...
+            if type(numbers) is str:
+                color = self.determine_color_of_number(numbers)
+                result = color(numbers)
+            else:
+                result = ", ".join(
+                    [
+                        self.determine_color_of_number(number)(number)
+                        for number in numbers
+                    ]
+                )
```
aconrad, line 126: Is this formatted by Black? Otherwise, I'd suggest having it more compact as such:

```suggestion
                result = ", ".join([
                    self.determine_color_of_number(number)(number)
                    for number in numbers
                ])
```

pycobertura/reporters.py
```diff
@@ -1,90 +1,64 @@
...
+env.filters["misses_color"] = filters.misses_color
 
-row_attributes = ["filename", "total_statements", "total_misses", "line_rate"]
-file_row = namedtuple("FileRow", row_attributes)
-file_row_missed = namedtuple("FileRowMissed", row_attributes + ["missed_lines"])
+headers_without_missing = ["Filename", "Stmts", "Miss", "Cover"]
+headers_with_missing = ["Filename", "Stmts", "Miss", "Cover", "Missing"]
 
 
 class Reporter(object):
```
aconrad, line 18: Since we only support py3 now, should we remove `object`?

```suggestion
class Reporter:
```

What do you think? If yes then we should do this in a separate PR as a follow-up to #123.

aconrad (commented):
I love this, the reports are really more concise. I left a few more comments, but overall this is looking great! Thanks for all your work on the reporters!

pycobertura/reporters.py
gro1m: it is indeed formatted by black :)

pycobertura/reporters.py
gro1m: I think that would be good :)

tests/test_stringify.py
gro1m: thanks!

pycobertura/reporters.py
gro1m: Yes, had to with your comment here: https://github.com/aconrad/pycobertura/pull/124#discussion_r784635006

pycobertura/templates/filters.py
gro1m: done

pycobertura/templates/html-delta.jinja2
gro1m: done -> renamed to `file_index`

pycobertura/reporters.py
gro1m: done

pycobertura/reporters.py
gro1m: done

pycobertura/reporters.py
gro1m: made a PR here: https://github.com/aconrad/pycobertura/pull/126. Also includes PRs from #123 (only the last 2 commits should matter)

gro1m: Ok that PR #126 looks more review-friendly now:)

pycobertura/reporters.py
gro1m: But I think we can still keep it or would you prefer to have `self.cobertura.files()` explicitly everywhere?

gro1m:
I will also do the CHANGES section again, once you are happy with the rest in this PR :)

pycobertura/reporters.py
aconrad: Oh wow, awesome, thanks!

pycobertura/reporters.py
aconrad: I'd prefer using `self.cobertura.files()`.

I see how my comment might've been confusing. When I talked about performance, I meant the overhead of the for-loop in Python since we're doing it a bunch of times now (vs. once previously). It didn't mean the cost of computing `self.cobertura.files()` (as I said above, this call is computed once and then memoized).

Don't worry about the performance part, I think it's negligible. Let's use `self.cobertura.files()` and we'll be good to merge this PR! 🎉

gro1m: done

aconrad:
Merged! 🎉

Thanks for all your work on this! Great clean-up of the `reporters.py` file. This lays the ground for easier-to-implement output formats!

PR #126, opened 2022-01-21
Maintentance/update to python37 follow up

gro1m:
remove object everywhere as new-style classes inherit by default from object in Python3 as discussed in PR #124


aconrad:
This is great, thank you! I'll merge it after your big PR is merged, I don't want to risk introducing conflicts on the big one.

aconrad:
As expected, this now conflicts but it should be trivial to fix. 👍

gro1m:
> As expected, this now conflicts after merging #124 but it should be trivial to fix. 👍

Yes, fixed :)

aconrad:
Thanks!!

PR #137, opened 2022-03-13
add possibility to hide columns

gro1m:
Add possibility to hide columns.

To investigate:
- Order of columns seems to be not anymore preserved :)
- consistency across all formats
- tests


gro1m:
@aconrad Let me know what you think of this implementation...

pycobertura/reporters.py
```diff
@@ -38,33 +39,40 @@ def get_report_lines(self):
...
-        lines["Missing"].append("")
+        footer_func_dict = {
+            "Stmts": [self.cobertura.total_statements()],
+            "Miss": [self.cobertura.total_misses()],
+            "Cover": [self.format_line_rate(self.cobertura.line_rate())],
+            "Missing": [""],
+        }
+        return rows_func_dict[key] + footer_func_dict[key]
+
+    def get_report_lines(self):
```
aconrad, line 57: I'd like to rename this to `get_summary_lines()` (and elsewhere when it applies), which is more self-explanatory. The wording "report" in this context is too generic because reports include the summary and source code lines. If you agree, I think we should also make other variables such as "lines" be renamed to "summary_lines", whenever that context applies. What do you think?

pycobertura/cli.py
```diff
@@ -66,6 +66,13 @@ def get_exit_code(differ, source):
     type=str,
     help="Delimiter for csv format, e.g. ,;\n\t",
 )
+@click.option(
+    "-hide",
+    "--hidecolumns",
```
aconrad, line 71: Can we call this

```suggestion
    "--hide-columns",
```

... ?

pycobertura/cli.py
```diff
@@ -89,7 +96,7 @@ def get_exit_code(differ, source):
     "the --source is a zip archive and the files were zipped under "
     "a directory prefix that is not part of the source.",
 )
-def show(cobertura_file, format, delimiter, output, source, source_prefix):
+def show(cobertura_file, format, delimiter, hidecolumns, output, source, source_prefix):
```
aconrad, line 99:
```suggestion
def show(cobertura_file, format, delimiter, hide_columns, output, source, source_prefix):
```

pycobertura/reporters.py
```diff
@@ -204,43 +250,28 @@ def get_report_lines(self):
...
+            + ["TOTAL"]
         }
 
-        lines["Filename"].append("TOTAL")
-        lines["Stmts"] += [
-            self.format_total_statements(self.differ.diff_total_statements())
-        ]
-        lines["Miss"] += [self.format_total_misses(self.differ.diff_total_misses())]
-        lines["Cover"] += [self.format_line_rate(self.differ.diff_line_rate())]
+        for key in set(headers_hideable).difference(set(self.hide_columns)):
```
aconrad, line 258: Can we store the output of `difference()` in a variable in order to label it?

```suggestion
        headers_to_show = set(headers_hideable).difference(set(self.hide_columns))
        for header_name in headers_to_show:
```

aconrad: We should be consistent on the naming used, "columns" vs. "headers". Can we pick one and be consistent throughout the code?

aconrad (commented):
Hi @gro1m, thanks for this PR.

I agree with the direction that PR is taking! I left a few comments and overall things are looking good. :)

pycobertura/cli.py
```diff
@@ -140,6 +148,13 @@ def show(cobertura_file, format, delimiter, output, source, source_prefix):
 )
 @click.argument("cobertura_file1")
 @click.argument("cobertura_file2")
+@click.option(
+    "-hide",
+    "--hidecolumns",
+    default="[]",
+    type=list,
```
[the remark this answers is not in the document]
gro1m: Resolved - thanks :)

gro1m:
Hi @aconrad 
I get file system not found errors, can you investigate this. In commit https://github.com/aconrad/pycobertura/pull/137/commits/a5ecd916b818ff9ab8a934b8bce1556cf67e36ef I could fix test_diff__better_coverage_has_exit_status_of_zero` by adjusting the paths in the xml, but I could not fix `test_diff__worse_coverage_exit_status` by adjusting paths. Do you have an idea what I need to do here to get this properly done? It just seems that all the other tests use two folders, e.g. 'tests/dummy.source1/coverage.xml' and only the dummy folder uses only one, e.g. 'tests/dummy.with-dummy2-no-cov.xml'. So I am not sure what to do here...

gro1m:
I needed to fix some tests here. Can you please check, @aconrad?

gro1m:
I needed to fix some tests here. Can you please check, @aconrad?

gro1m:
Closed in favor of https://github.com/aconrad/pycobertura/pull/160

PR #132, opened 2022-02-05
Fix/handle multiple classes in same files correctly

gro1m:
Hi @aconrad 
I try to continue on PR #113 here, as @smortvedt did not yet reply and may have lost interest in the project.

As a first, I just merged in his changes into my master and put it into a new branch.

PR #138, opened 2022-03-13
add regex option

gro1m:
Adds a regex option.

Examples:

```
$ pycobertura show coverage.xml --regex "^c"
Filename        Stmts    Miss  Cover    Missing
------------  -------  ------  -------  ---------
cli.py             78       0  100.00%
cobertura.py      206       0  100.00%

pycobertura show coverage.xml --regex "^u"
Filename      Stmts    Miss  Cover    Missing
----------  -------  ------  -------  ---------
utils.py        118       1  99.15%   38
TOTAL           695       1  99.86%
```

Did not test diff yet, there are also no tests yet...

Just wanted to get your opinion on it, @aconrad. I tried to keep the logic simple and here it is for including the files you want but you could also write the regex as such as to exclude the files you do not want. It is a bit complicated, but it works:
```
pycobertura show coverage.xml --regex '^(?!u|\\.).*'
Filename                 Stmts    Miss  Cover    Missing
---------------------  -------  ------  -------  ---------
reporters.py               192       0  100.00%
__init__.py                  2       0  100.00%
filesystem.py               87       0  100.00%
templates/__init__.py        0       0  100.00%
cobertura.py               206       0  100.00%
templates/filters.py        12       0  100.00%
cli.py                      78       0  100.00%
TOTAL                      695       1  99.86%
```
The single quotes are absolutely necessary, otherwise one gets: 
```bash
bash: !u: event not found
```


aconrad:
Can you update from master? I enabled Github Actions although we don't have the coverage pipeline working for now, will have to do that subsequently. See #139.

gro1m:
> Can you update from master? I enabled Github Actions although we don't have the coverage pipeline working for now, will have to do that subsequently. See #139.

Yes, done right now and at least the checks seem to be back again and I think they should fail :)

gro1m:
@aconrad The tests fail, because the Main class is shown twice. Indeed, the Main class occurs twice in cobertura.xml, so is the test actually correct and should we modify the assertions. Could you please check?

aconrad:
Hm, this `cobertura.xml` file was taken from the internet somewhere, I didn't generate it – I was looking for a sample cobertura file to use for testing when I started pycobertura. But I did a git blame on it and noticed that this file was tampered with in 1fb4d0b1 (PR https://github.com/aconrad/pycobertura/pull/82), was a bug introduced there?

gro1m:
Hi @aconrad
Thanks for checking! 
I think for now it would be best to remove these lines (maybe you could open a separate PR so that we have this well tracked). And in future it might be good if we could generate these files ourselves, but I think this would be an enhancement that we might not want to do right now...
What do you think?

aconrad:
> I think for now it would be best to remove these lines (maybe you could open a separate PR so that we have this well tracked).

I'll have to take a closer look at why we have a filename that doesn't look like a path, I'm not familiar with this syntax. Is it a Java thing? We can follow up in the original PR and ask the author.

> And in future it might be good if we could generate these files ourselves, but I think this would be an enhancement that we might not want to do right now...

I didn't want to generate my own to ensure that pycobertura wasn't biased parsing a non standard file format (e.g. as you noticed, even the coveragepy-generated files are broken by omitting the directory in the filename attributes). The original coverage schema standard is not very precise in describing what the value of each attribute should be.

> What do you think?

I would encourage introducing files generated by other languages. But if we were to generate our own, how would we go about it? Do you mean the ones generated by the pycobertura project? What value would it bring vs the ones we already have? Some of the sample coverage files in the tests directory were generated ourselves.

gro1m:
> > I think for now it would be best to remove these lines (maybe you could open a separate PR so that we have this well tracked).
> 
> I'll have to take a closer look at why we have a filename that doesn't look like a path, I'm not familiar with this syntax. Is it a Java thing? We can follow up in the original PR and ask the author.
> 
> > And in future it might be good if we could generate these files ourselves, but I think this would be an enhancement that we might not want to do right now...
> 
> I didn't want to generate my own to ensure that pycobertura wasn't biased parsing a non standard file format (e.g. as you noticed, even the coveragepy-generated files are broken by omitting the directory in the filename attributes). The original coverage schema standard is not very precise in describing what the value of each attribute should be.
> 
> > What do you think?
> 
> I would encourage introducing files generated by other languages. But if we were to generate our own, how would we go about it? Do you mean the ones generated by the pycobertura project? What value would it bring vs the ones we already have? Some of the sample coverage files in the tests directory were generated ourselves.

Actually we would need to test some popular coverage report generators always on the same code logic but in different languages. Yes, I agree this would be challenging and I currently have no idea on how to go about this, but maybe someone else from the community could support here.
Also the code logic would have to be small, but still relevant.
The only reason I brought it up is that if the truth lies in the project's code, it is much easier to understand the code and why certain changes have evolved over time.

gro1m:
@aconrad I think I still need your decision here about the `cobertura.xml`, i.e. correcting the tests to account for the duplicate `Main.java` or removing the corresponding line from the `cobertura.xml`...

aconrad:
The method `Cobertura.files()` was designed to deduplicate filenames if they appeared multiple times and we should retain that behavior since that seems to be a valid use case (the motivation for #82).

There can be multiple classes in the same filename, and since the `<class>` element is meant to represent a class object, it's possible to find a different class within the same file. It makes sense to me.

gro1m:
> The method `Cobertura.files()` was designed to deduplicate filenames if they appeared multiple times and we should retain that behavior since that seems to be a valid use case (the motivation for #82).
> 
> There can be multiple classes in the same filename, and since the `<class>` element is meant to represent a class object, it's possible to find a different class within the same file. It makes sense to me.

Ok, but should we then not include all of them, i.e. does it not work against https://github.com/aconrad/pycobertura/pull/132.

For now, I reverted the logic so that this PR should be good to review, but I am not sure I quite understood the above...

aconrad:
> but I am not sure I quite understood the above...

My understanding is that if we have a file `foods.py`, this file could have `class Apple: ...` and `class Banana: ...`. The coverage tool could generate the following report:

```xml
<coverage>
  <packages>
    <package ...>
      <classes>
        <class name="Apple" filename="foods.py" ...>
          <methods>...</methods>
        </class>
        <class name="Banana" filename="foods.py" ...>
          <methods>...</methods>
        </class>
      </classes>
    </package>
  </packages>
</coverage>
```

When we call `Cobertura.files()`, the expectation is that `["foods.py"]` would be returned, not `["foods.py", "foods.py"]`. This is why #82 was implemented.

Does it add clarity?

aconrad:
Hm. Let me take a closer look at #132, I haven't looked into it yet.

gro1m:
Hi @aconrad 
After thinking more about it now, I understood that the current implementation is indeed correct, because we measure the stats on a file-based level. The only question one might ask is why are we parsing on `\\class` and not on something like `file`, as we have file-based stats and not class-based stats: is this not possible within `lxml`? And also one could discuss if we should offer different views, e.g. directory level, file level, class level, method level, maybe even a tree. But I think the discussion now diverges from what the PR additionally wanted to address and should probably take place in a different issue, if you think it could be worth it to discuss this further.

aconrad:
> The only question one might ask is why are we parsing on `\\class` and not on something like `filename` or `file path`(because there could be 2 files with same name in different directories), as we have file-based stats and not class-based stats: is this not possible within `lxml`?

Great question. When I intially implemented pycobertura, I used coverage reports generated by coveragepy as my sample coverage files. And we can tell from the report that 1 class == 1 file. Here's a snippet coverage from our `tests/dummy.original.xml` directory:

```xml
<class branch-rate="0" complexity="0" filename="dummy/__init__.py" line-rate="0" name="dummy/__init__">
        <methods/>
        <lines/>
</class>
<class branch-rate="0" complexity="0" filename="dummy/dummy.py" line-rate="0.5" name="dummy/dummy">
        <methods/>
        <lines>
                <line hits="1" number="1"/>
                <line hits="0" number="2"/>
                <line hits="1" number="4"/>
                <line hits="0" number="5"/>
        </lines>
</class>
```

So that's why the implementation is like that.

But over time, people have reported issues because their coverage files (generated by other tools/languages) aren't properly parsed by pycobertura (e.g. #82). Pycobertura is meant to be a [Cobertura](https://github.com/cobertura/cobertura) file parser and be language agnostic.

> And also one could discuss if we should offer different views, e.g. directory level, file level, class level, method level, maybe even a tree. But I think the discussion now diverges from what the PR additionally wanted to address and should probably take place in a different issue, if you think it could be worth it to discuss this further.

I agree.

pycobertura/utils.py
```diff
@@ -229,3 +231,29 @@ def hunkify_lines(lines, context=3):
...
+    with open(file_path, "rb") as f:  # read in binary (more secure)
+        result = [
+            line.decode("ascii").strip()
+            for line in f.readlines()
+            if not line.decode("ascii").startswith(comment_character)
+        ]
+        return [res for res in result if res != ""]
+
+
+def get_filenames_that_do_not_match_regex(
```
aconrad, line 246: Could we use a tool like this one? https://github.com/mherrmann/gitignore_parser – would it have the same effect?

gro1m: The effect would be similar (it though does only handle files), but this library did not have any changes since August 2020 and my implementation seems to use more Python in-built functions. Also the former does not have a stable semantic version (0.0.8). In priniciple, best practice is to only include well-maintained libraries and such that have a major version >= 1.

gro1m: @aconrad Maybe this library could be used: https://github.com/bitranox/igittigitt. And also tell me if you have any other things still needed here to get into a good state to merge - thanks :)

aconrad:
This looks good, thanks! Can we update CHANGES file? Also, should we add a note somewhere in the README about this new option?

gro1m:
> This looks good, thanks! Can we update CHANGES file? Also, should we add a note somewhere in the README about this new option?

done, can you please check again :)

README.md
```diff
@@ -319,6 +319,48 @@ pycobertura diff --format yaml tests/dummy.source1/coverage.xml tests/dummy.sour
 
 ![Example output of yaml formatted pycobertura diff command](images/example_yaml_output.png)
 
+### Option to exclude files via --ignore-regex-option
+You can specify the --ignore-regex option to exclude files that have a certain pattern either by specifying the path to a .gitignore file or by entering a Python regex.
```
aconrad, line 323:
```suggestion
You can specify the `--ignore-regex` option to exclude files that have a certain pattern either by specifying the path to a `.gitignore` file or by entering a Python regex.
```

README.md
```diff
@@ -319,6 +319,48 @@ pycobertura diff --format yaml tests/dummy.source1/coverage.xml tests/dummy.sour
...
+```
+```
+pycobertura show tests/cobertura.xml --ignore-regex ".*Search"
+Filename      Stmts    Miss  Cover    Missing
+----------  -------  ------  -------  ---------
+Main.java        15       0  100.00%
+TOTAL            34       3  90.00%
+```
+> NOTE: Putting just a *, e.g. "*Search" above will lead to an error, you need to really follow Python regex conventions to use this option.
+You can also specify a .testgitignore with the following content:
```
aconrad, line 343:
```suggestion
You can also specify a `.testgitignore` with the following content:
```

CHANGES.md
```diff
@@ -25,6 +25,7 @@
     `json`, `markdown`, `csv`, and `yaml`.
   * Only uncovered lines are reported in `Missing` column (instead of
     additionally reporting newly covered lines)
+  * Add option `--ignore-regex` to ignore some files using either a Python regex pattern or a path to a .gitignore file.
```
aconrad, line 28:
```suggestion
  * Add option `--ignore-regex` to ignore some files using either a Python regex pattern or a path to a `.gitignore` file.
```

README.md
```diff
@@ -319,6 +319,48 @@ pycobertura diff --format yaml tests/dummy.source1/coverage.xml tests/dummy.sour
...
+TOTAL                                34       3  90.00%
+```
+```
+pycobertura show tests/cobertura.xml --ignore-regex ".*Search"
+Filename      Stmts    Miss  Cover    Missing
+----------  -------  ------  -------  ---------
+Main.java        15       0  100.00%
+TOTAL            34       3  90.00%
+```
+> NOTE: Putting just a *, e.g. "*Search" above will lead to an error, you need to really follow Python regex conventions to use this option.
```
aconrad, line 342:
```suggestion
> NOTE: Putting just a `*`, e.g. "*Search" above will lead to an error, you need to really follow Python regex conventions to use this option.
```

aconrad (approved):
Looks great, thanks! I suggested some formatting but that's about it! Ready to merge! Thank you!

PR #141, opened 2022-03-26
move test requirements into setup.cfg

gro1m:
Move tests into setup.cfg. Would also allow to pin version of `pytest`.


gro1m:
@aconrad Do not merge before #140 is merged and this needs some merge conflict resolution, especially, the `pytest-cov` package must disappear.

aconrad:
Do we still need/want this PR @gro1m?

aconrad:
Isn't the industry standard to use `test-requirements.txt`? What motivates this change?

gro1m:
See the discussion here (there are preferences for either, but maybe it can help you form an opinion on this): https://discuss.python.org/t/is-it-preferable-to-add-test-dependensies-as-an-extra-require-instead-of-tox-ini/4984.

aconrad:
Typing `pip install -e .[test]` feels a little more cryptic to me, but it's likely just about habit. But I haven't worked professionally with Python for a couple of years and if that's how people like to do it nowadays, then fine by me.

aconrad:
The PR description says that it would "allow to pin version of `pytest`". Why was that not already the case with `test-requirements.txt`?

gro1m:
> The PR description says that it would "allow to pin version of `pytest`". Why was that not already the case with `test-requirements.txt`?

I am not sure why I put that. At the time, I thought that would be a particular advantage, but it is also possible with `test-requirements.txt`, as I did in my last PR.

gro1m:
But I think the discussion is more if you want to keep the test requirements external to the package or not. And what is nicer with the `setup.cfg` is that you do not explicitly specify the current package with a `.`

PR #151, opened 2022-06-03
Coverage parsing speedup

oev81:


setup.cfg
```diff
@@ -40,6 +39,9 @@ install_requires =
     tabulate
     ruamel.yaml
 
+[options.package_data]
+* = *.jinja2, *.css
```
oev81, line 43: Without this change, installing from archive misses the package data files.

gro1m:
@aconrad Could you please review? This looks helpful...

setup.cfg
aconrad: @gro1m I'm less familiar with this part, any concerns?

aconrad (approved):
This looks great, thank you for your contribution @oev81!

The `black` command is failing, I made a change so the console output can tell us what's wrong instead of just failing without further information. https://github.com/aconrad/pycobertura/pull/152

aconrad:
Can you pull the lastest master into your branch so we can see what the `black` issue is?

aconrad:
Can you also update the CHANGES file with a note about the fix? Feel free to tag yourself (see other entries as an example)

setup.cfg
aconrad: @oev81 what did you observe? What do you mean by "installing from archive"? Do you mean from a zip file instead of `pip install pycobertura`? What was your process to get an archive?

gro1m: > @gro1m I'm less familiar with this part, any concerns?

This is about data file support.
In this case all `jinja2` and `css` files are included as well and not only the Python files. 
Concretely, this means when installing pycobertura, we have also access to the following files in the `templates` folder:
- `html.jinja2`
- `html-delta.jinja2`
- `normalize.css`

As we have this under control, this should not be a security issue, but rather an improvement. Probably some html rendering might not have worked before in all cases, as for `include_package_date=True` to have worked properly, an additional `MANIFEST.in` file (see also the reference link below) would have been required.

Data File Support Reference (even if for `setup.py` instead of `setup.cfg`, but this analogue): https://setuptools.pypa.io/en/latest/userguide/datafiles.html

Hope this is clearer now.

oev81: @aconrad 
You can install package directly from github as an archive
```
pip install https://github.com/oev81/pycobertura/archive/coverage_parsing_speedup.zip
```
or put
```
https://github.com/oev81/pycobertura/archive/coverage_parsing_speedup.zip
```
in requirements.txt

This way is very helpful for me.  https://stackoverflow.com/a/24811490

CHANGES.md
```diff
@@ -20,6 +20,9 @@
 * Replace Travis-ci with Github Actions. We were having issues with Travis
   pipeline not triggering because of quota issues.
 
+* Changes in Cobertura class. All `"./packages//class"` elements extracted at once now, and this leads to significant speedup.
+* Changes in setup.cfg related to package data. Remove `include_package_data=True`, because it requires `MANIFEST.in` and add `[options.package_data]` instead. This change makes the package installing from archive not miss `*.css` and `*.jinja2` files.
```
aconrad, line 24: This is some pretty sweet changelog, thank you for doing that!

PR #152, opened 2022-06-03
Show diff output of black when it needs changes

aconrad:
When the `black` fails, this will show what the diff would be.


aconrad:
cc @gro1m

Issue #146, opened 2022-04-11
Export JSON as a list of objects


notpushkin:
First of all thanks for your wonderful project! 🎉 

Current JSON layout (column-oriented) isn't really easy to consume. I propose we flip it like this:

```js
{
  "files": [
    {
      "Filename": "pycobertura/__init__.py",
      "Stmts": 2,
      "Miss": 0 ,
      "Cover": "100.00%",
      "Missing": ""
    },
    // ...
  ]
}
```

As a sidenote, I'd also like to shout out an excellent library called [tablib](https://tablib.readthedocs.io/), which [supports a ton of formats](https://tablib.readthedocs.io/en/stable/formats.html) out of the box and would probably be easier to use than maintaining a lot of different formatters by hand.

notpushkin:
P. S. same applies to Yaml, of course.

gro1m:
@notpushkin @aconrad 
I think the format is totally feasible, with regards to the library I do not have yet an opinion on it, as I do not know it. Generally, from a maintenance perspective I think in-built solutions and as less external packages as possible is optimal, so maybe in future a decision between `tabulate` and `tablib` would have to be taken.

Regarding implementation: I am currently still looking at the pytest coverage issue, which is a release-blocker for Version 3.0 and the solution could be to switch to `unittest`, so that is maybe just a `caveat` that it might be good to come up with something small first :)

gro1m:
@notpushkin How do you get along with this PR? It seems the changes the resolution of the pytest coverage issue brings will be minor, if this time I really have found the solution. For more info please look at: #140.

aconrad:
I'm marking this issue for milestone 3.0 since it would be a breaking change.

aconrad:
pycobertura-3.0.0 was released, I hope this release addresses this issue. Please reopen if not.

PR #153, opened 2022-06-05
Export Json and Yaml as List of Objects

gro1m:
Addresses https://github.com/aconrad/pycobertura/issues/146.

@aconrad Please review!
FYI: @notpushkin


pycobertura/reporters.py
```diff
@@ -54,6 +54,16 @@ def get_report_lines(self):
 
         return lines
 
+    def per_file_stats(self, lines):
+        rows = []
+        items = lines.items()
+        number_of_files = len(self.cobertura.files())
+        for i in range(number_of_files):
+            if i < number_of_files:
```
notpushkin, line 62: Won't `i` always be less than `number_of_files` here? 🤔

gro1m: Should be addressed in 0f5035c81f0021caaaffa9e953480d7f00c76f81. Though what confuses me a little is that for the diff the formula is:
```
number_of_files = len(self.differ.files()) - 1
```

, whereas for the plain show the formula is:

```
number_of_files = len(self.cobertura.files())
```

Note the missing -1 in the second case.
So could one of you, @aconrad or @notpushkin, check that this is really correct.

pycobertura/reporters.py
```diff
@@ -54,6 +54,13 @@ def get_report_lines(self):
 
         return lines
 
+    def per_file_stats(self, lines):
```
aconrad, line 57: The reading of this function is a little cryptic. Any chance we can replace `k` and `v` with more descriptive names? Maybe `k` could be `header_name` or something like that?

Also, would it be helpful to have a docstring that illustrates what will be returned by this function so readers understand what to expect?

aconrad: I like to use an ellipsis when it makes sense to avoid being too verbose, such as:

```python
def per_file_stats(self, lines):
    """
    Returns dict with keys `files` and `total` that contain coverage
    statistics per file and overall, respectively.

    {
        "files": [
            {
                "Filename": "dummy/__init__.py",
                "Stmts": 0,
                "Miss": 0,
                "Cover": "0.00%",
                "Missing": "9"
            },
            ...
        ],
        "total": {
            "Filename": "TOTAL",
            "Stmts": "...",
            "Miss": "...",
            "Cover": "..."
        }
    }
    """
    # <... code ...>
```

What do you think?

aconrad: Also, I'd like to rename `lines` to something less ambiguous. `lines` here is the output of `get_report_lines` which is quite different than `line` used elsewhere in the code base.

aconrad:
Ah, we're getting conflicts now, naturally.

pycobertura/reporters.py
gro1m: done in recent commit 1c6d2da62c48fb443daf4d3c9350787fc23f1a5a.

gro1m:
> Ah, we're getting conflicts now, naturally.

resolved :)

pycobertura/reporters.py
```diff
@@ -246,6 +283,13 @@ def get_report_lines(self):
...
+        items = lines.items()
+        number_of_files = len(self.differ.files()) - 1
+        rows = [{k: v[i] for k, v in items} for i in range(number_of_files)]
+        footer = {k: v[number_of_files] for k, v in items if k != "Missing"}
+        return {"files": rows, "total": footer}
```
aconrad, line 291: Can we apply the same variable naming convention as you used in the other function? The changes you did above look great.

aconrad (approved):
Looking good! Can we update the README with the YAML/JSON samples to match the new output format?

pycobertura/reporters.py
gro1m: updated both functions a little and used same naming convention. Please re-review!

pycobertura/reporters.py
```diff
@@ -54,6 +54,45 @@ def get_report_lines(self):
...
+        file_stats_dict_items = file_stats_dict.items()
+        number_of_files = len(self.cobertura.files()) + 1
+        file_stats_list = [
+            {
+                header_name: header_value[file_index]
+                for header_name, header_value in file_stats_dict_items
+            }
+            for file_index in range(number_of_files)
+        ]
+        file_stats_list[-1].pop("Missing", "No key Missing found")
```
aconrad, line 90:
```suggestion
        file_stats_list[-1].pop("Missing", 'Key "Missing" not found')
```

pycobertura/reporters.py
```diff
@@ -246,6 +282,22 @@ def get_report_lines(self):
...
+        file_stats_dict_items = file_stats_dict.items()
+        number_of_files = len(self.differ.files())
+        file_stats_list = [
+            {
+                header_name: header_value[file_index]
+                for header_name, header_value in file_stats_dict_items
+            }
+            for file_index in range(number_of_files)
+        ]
+        file_stats_list[-1].pop("Missing", "No key Missing found")
```
aconrad, line 295:
```suggestion
        file_stats_list[-1].pop("Missing", 'Key "Missing" not found')
```

aconrad (approved):
This looks good. We need to resolve that conflict first, though.

aconrad:
@gro1m would you want me to take over this PR or can you fix the merge conflict so we can get this one merged?

gro1m:
Hi @aconrad 
I tried to merge the PR and resolve the conflict, but currently I can only do this via the editor, which is a bit error-prone. So if you could resolve it, great. Otherwise it will take me certainly another week to have a look at it.

aconrad:
Sounds good. I'll try to take a look at it today.

aconrad:
Oh, you did merge master and resolve the conflict actually. I can look at the failing tests then.

aconrad:
@gro1m here's a PR to fix tests: https://github.com/gro1m/pycobertura/pull/1

It's a PR against your branch on your fork. Not sure if that's the best way to go about it but heh, let's give it a try. 😄

gro1m:
> @gro1m here's a PR to fix tests: [gro1m#1](https://github.com/gro1m/pycobertura/pull/1)
> 
> It's a PR against your branch on your fork. Not sure if that's the best way to go about it but heh, let's give it a try. 😄

Hi @aconrad 
That worked well, thanks! :)

notpushkin:
🎉 
Thanks folks!

Issue #148, opened 2022-04-18
"Missing" column shouldn't include covered lines


aconrad:
Currently, the "Missing" column shows uncovered lines introduced as well as newly covered ones. Showing covered lines under "Missing" is confusing because developers care about uncovered lines and, well, if it's covered then it's not missing. 😏

See [original comment](https://github.com/aconrad/pycobertura/issues/142#issuecomment-1092310811).

aconrad:
pycobertura 3.0.0 was released

PR #154, opened 2022-06-05
only return uncovered lines in Missing column

gro1m:
Addresses https://github.com/aconrad/pycobertura/issues/148.
@aconrad Please review!
Unfortunately the `file_source` function is a bit unreadable for me, so I chose to return a tuple where the second element is always True. But probably at some point, we would like to separate the `line-edit` part.

### DIFF
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -366,7 +366,7 @@ class CoberturaDiff:
         The given file `filename` where `lineno` is a missed line number.
         """
         return [
-            (line.number, True)
+            line.number
             for line in self.file_source(filename)
             if line.status is False
         ]
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -362,17 +362,10 @@ class CoberturaDiff:
 
     def diff_missed_lines(self, filename):
         """
-        Return a list of 2-element tuples `(lineno, is_new)` for the given
-        file `filename` where `lineno` is a missed line number and `is_new`
-        indicates whether the missed line was introduced (True) or removed
-        (False).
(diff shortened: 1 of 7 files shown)


pycobertura/cobertura.py
```diff
@@ -362,17 +362,14 @@ def diff_line_rate(self, filename=None):
...
-                line_changed.append((line.number, is_new))
-        return line_changed
+        Return a list of 2-element tuples `(lineno, True)` for uncovered lines.
+        The given file `filename` where `lineno` is a missed line number.
+        """
+        return [
+            (line.number, True)
+            for line in self.file_source(filename)
+            if line.status is False
+        ]
```
aconrad, line 372: The second item returned in the tuple helped reporters colorize added/remove lines (green/red). Although, since we're deciding that the missed column will only contain missed lines, there isn't a need anymore to return that second item in the tuple. We can actually return a scalar value of the missed line number only.

```suggestion
        return [
            line.number
            for line in self.file_source(filename)
            if line.status is False
        ]
```

Then we'd have to update the other parts of the code that depend on `diff_missed_lines()` to not treat the value as a tuple anymore.

Additionally, I don't think we need to render the `+` anymore (see `format_missed_lines()`) since we no longer need to distinguish an uncovered line from a covered one anymore. Although I think renderers should still render missed lines as red as it highlights to the user where their next coverage efforts should be put into.

Does that make sense?

aconrad (approved):
Approved! This is great, thank you! Can you please add an entry in CHANGES about this change and then I'll merge.

aconrad:
Wonderful, thank you so much @gro1m !!

Issue #144, opened 2022-04-08
[Feature Request] Do a delta on previously non-exist files


chixiangzhou:
Adding a new file is a common use case in software development. Developers might be interested in how many lines are covered in the new file, and what is the delta.

Currently, pycobertura diff report considers that previously non-exist files cannot do a delta, and marks the Column **Cover** as "-".  IMO, it would make more sense to consider that previously non-exist files are "100%" covered, and do a delta after the new file is introduced. Thus, the diff coverage can be shown in the Column **Cover.**

aconrad:
This should've been addressed with https://github.com/aconrad/pycobertura/pull/155

aconrad:
pycobertura-3.0.0 was released, I hope this release addresses this issue. Please reopen if not.

PR #155, opened 2022-06-11
100% or 0 for non-available stats

gro1m:
Addresses https://github.com/aconrad/pycobertura/issues/144.
Can you please review, @aconrad ?


aconrad:
Thanks for this! I'm sorry it took so long to review.

PR #159, opened 2022-10-09
Cache coverage reports against master

aconrad:
WIP: currently using Github Action [cache](https://github.com/actions/cache) to store coverage artifacts of `master` branch. But artifacts are only accessible for a week.

> A repository can have up to 10GB of caches. Once the 10GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.

[Source](https://github.com/actions/cache#cache-limits).

Given how infrequent the development against `master` is, the coverage generated by the `master` branch is unlikely to be available the next time a new PR is opened and coverage diff won't be able to work.


goatwu1993:
I had setup a similar workflow, PR comment included(if anyone is interested...)

.github/workflows/build-and-test-pycobertura.yml
```diff
@@ -14,16 +14,26 @@ jobs:
...
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           python -m pip install tox tox-gh-actions
+
       - name: Test with tox
         run: tox
+        
+      - name: Upload coverage artifact
+        if: ${{ env.GITHUB_REF_NAME == 'master' }}
```
goatwu1993, line 35: if: ${{ github.event_name == 'push' }}

Issue #107, opened 2020-11-24
Bug using ZipFileSystem and DirectoryFileSystem together


chezihp:
Hey,
Thanks for this project. Really helpful!
Im using both ZipFileSystem and DirectoryFileSystem together (one for Cobertura1 and the other for Cobertura2). Found out that the line mapping between the two FileSystems is not working properly because ZipFileSystem adds 'b' prefix to each source line (this is [expected](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open)) while DirectoryFileSystem doesn't. This effects cases for example when the same source line is missed in both coverage.xml files but the line still marked as status=False and reason="line-edit". 
The workaround Im currently using is to override the ZipFileSystem.open method and wrapping the opened file with TextIOWrapper.

aconrad:
Thanks for your report @chezihp, and nice find! Do you get a traceback, on just incorrect output from pycobertura?

Is there any chance for you to put a PR together with a test that displays the failure?

chezihp:
Hey.
Just incorrect output. I'll put a PR as soon I'll find the time to do it.

aconrad:
Can you provide a reproducible test case?

ernestask:
[pycobertura.zip](https://github.com/aconrad/pycobertura/files/11613554/pycobertura.zip)

```bash
➜  pycobertura pycobertura diff --source1 src.zip --source2 . cobertura1.xml cobertura2.xml
Filename      Stmts    Miss  Cover     Missing
----------  -------  ------  --------  ---------
TOTAL             0       0  +100.00%
➜  pycobertura echo $?
3

➜  pycobertura pycobertura diff --source1 . --source2 . cobertura1.xml cobertura2.xml
Filename      Stmts    Miss  Cover     Missing
----------  -------  ------  --------  ---------
TOTAL             0       0  +100.00%
➜  pycobertura echo $?
0
```

PR #169, opened 2023-05-31
filesystem: Wrap archived files inside TextIOWrapper

ernestask:
Currently, when mixing DirectoryFileSystem and ZipFileSystem, diffing reports wrong results for identical coverage reports due to differing types in file systems (byte-string lines compared with text lines).

This commit wraps the files opened from a zip archive inside TextIOWrapper to avoid the issue.

Closes https://github.com/aconrad/pycobertura/issues/107

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 skipsdist = True
-envlist = py37, py38, py39, py310, pep8, black
+envlist = py37, py38, py39, py310, py311, pep8, black
 
 [gh-actions]
 python =
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ python =
     3.8: py38
     3.9: py39
     3.10: py310, pep8, black
+    3.11: py311
 
 [flake8]
 max-line-length = 88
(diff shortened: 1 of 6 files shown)


aconrad:
Thanks for tackling this issue @ernestask, much appreciated. 🙏

* Is there a way we can demonstrate this fix in `tests/test_filesystem.py`?
* How about `tests/test_cobertura_diff.py`? I see you already posted a ZIP file that includes a reproducible scenario for this issue, thank you! Could we use that in the `tests/` directory to demonstrate the issue for a diff test?

ernestask:
Hey, yeah, I’ll look into that tomorrow.

aconrad:
Additionally, can we add an entry under `## Unreleased` in `CHANGES.md` that explains what was fixed?

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 skipsdist = True
-envlist = py37, py38, py39, py310, pep8, black
+envlist = py37, py38, py39, py310, py311, pep8, black
```
aconrad, line 3: I think we also need to update the section `[gh-actions]` below.

ernestask: Right, and the Actions workflow.

ernestask:
:pray:

Do you think .1 could be tagged with this? I’d start using it right away in that case.

aconrad:
https://pypi.org/project/pycobertura/3.2.1/

ernestask:
Awesome, thanks a lot!

PR #172, opened 2023-10-05
Account for moved/renamed/delted files in miss counts

wdouglass:
When files are renamed, pycobertura may falsely indicate an increase in missed lines, because deleted files aren't accounted for in the count. This patch rectifies that situation

Please let me know if I've made any bad assumptions, and thank you very much for this useful tool!


aconrad:
Hi @wdouglass,

Thanks for your contribution to pycobertura!

Could you take a moment to elaborate and illustrate with a few scenarios of what you observe that isn't right and what you expect to see? It would be helpful for me so I can better review this PR with your scenarios in mind. If you haven't already, adding a test specifically for your scenario would help too.

Thanks!

wdouglass:
I gladly will! So here's the problem I saw

Imagine I have a program, that's made of source files `main.c` and `a.c`. say `a.c` has 5 missed lines. Now, if `a.c` got renamed to `b.c`, a pycobura diff should show a zero delta in total missed lines, but it was counting the new missed lines in `b.c` without subtracting the lines that were no longer missed in `a.c` (which no longer exists), and showing +5 missed lines

I'll definitely generate some reports and write a test that exercises this tomorrow morning, and push it to this PR. Thanks again for maintaining this project!

wdouglass:
note: i force-pushed to add my GPG signature

tests/test_cobertura_diff.py
```diff
@@ -44,8 +44,16 @@ def test_diff_total_misses():
...
+    assert differ.diff_total_misses() == -4
 
+def test_diff_total_misses_move():
+    from pycobertura.cobertura import CoberturaDiff
+
+    cobertura1 = make_cobertura('tests/dummy.source1/coverage.xml')
+    cobertura2 = make_cobertura('tests/dummy.moved/coverage.xml')
+    differ = CoberturaDiff(cobertura1, cobertura2)
+
+    assert differ.diff_total_misses() == 0
```
aconrad, line 56: I think this test makes sense and illustrates well the problem you are solving. Thanks!

pycobertura/cobertura.py
```diff
@@ -376,27 +376,47 @@ def diff_missed_lines(self, filename):
...
+            lines2 = self.cobertura2.source_lines(filename)
+            line_statuses2 = dict(self.cobertura2.line_statuses(filename))
+            nonexistent = False
+        else:
+            lines2 = []
+            line_statuses2 = {}
+
+        if nonexistent:
+            # try to get source lines anyway, to get the exception traceback
+            self.cobertura2.source_lines(filename)
```
aconrad, line 419: Why is getting the exception useful here? Can you clarify?

wdouglass: I guess i wanted the exception in the case that the file doesn't exist in the filesystem of either cobertura1 or cobertura2. i guess the following code would be equivalent (but i'm not sure if it's more clear)
```suggestion
            self.cobertura2._raise_MissingFileSystem(filename)
```

wdouglass: Maybe i got tripped up by the fact that `MissingFileSystem` is a sub-class inside the `Cobertura` class

aconrad: It's fine. I looked at the code a little closer and what you did makes sense. Thanks!

aconrad (approved):

aconrad:
Released as v3.2.3.

PR #173, opened 2023-10-08
Fix: Number of files in delta reporter

ulasozguler:
When number of files with coverage change is less then total number of files in coverage report, `DeltaReporter.per_file_stats` throws a "IndexError: list index out of range" exception. 

In test case, both coverage files have 3 items. One of the items in `new-coverage.xml` has coverage change.


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `pycobertura diff` bug related to number of changed files. Thanks @ulasozguler
```
aconrad, line 5: Can we be more specific? Maybe reuse the wording used in the PR's description.

ulasozguler: Done.

aconrad (approved):
Thanks for fixing this! Much appreciated!

aconrad:
Released as v3.2.2

PR #174, opened 2023-12-07
Return None if branch-rate is absent

starcruiseromega:
If "branch-rate" isn't present in the XML, this method tries to return `float(None)`. I believe that branch-rate isn't required, at least by the "loose" Cobertura spec, but either way it seems more useful to be less strict.

### DIFF
--- a/pycobertura/cobertura.py
+++ b/pycobertura/cobertura.py
@@ -123,11 +123,13 @@ class Cobertura:
         Return the global branch rate of the coverage report. If the
         `filename` file is given, return the branch rate of the file.
         """
+        branch_rate = None
         if filename is None:
-            return float(self.xml.get("branch-rate"))
+            branch_rate = self.xml.get("branch-rate")
         else:
             classElement = self._class_elements_by_file_name[filename][0]
-            return float(classElement.get("branch-rate"))
+            branch_rate = classElement.get("branch-rate")
+        return None if branch_rate is None else float(branch_rate)
 
     @memoize
     def missed_statements(self, filename):


tests/test_cobertura.py
```diff
@@ -33,6 +33,10 @@ def test_parse_string():
         xml_string = f.read()
     assert ET.tostring(Cobertura(xml_path).xml) == ET.tostring(Cobertura(xml_string).xml)
 
+def test_no_branch_rate():
+    from pycobertura import Cobertura
+
+    assert Cobertura('tests/cobertura-no-branch-rate.xml').branch_rate() == None
```
aconrad, line 39: Can we move this test close to `test_branch_rate` below?

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Fix: Cobertura.branch_rate() returns None when branch rate is absent instead of raising TypeError
```
aconrad, line 5:
```suggestion
* Fix: `Cobertura.branch_rate()` returns `None` when `branch-rate` is absent from the Cobertura coverage report instead of raising `TypeError`. Thanks @starcruiseromega
```

aconrad (approved):

starcruiseromega:
Hi @aconrad ! Thanks for the merge and release, and sorry for the delay answering your question. We have an internal tool that tracks code coverage, and we use pycobertura to parse out the line coverage and branch coverage from users' reports. Most of them are generated by tools like pytest, but there are a few using languages like OCaml, for which branch coverage isn't very meaningful, so they don't have it in their reports.

PR #188, opened 2024-06-29
Add phorge-json format

tusooa:


tests/test_reporters.py
```diff
@@ -810,3 +810,41 @@ def test_delta_reporter__single_file_coverage_changed():
...
+def test_phorge_json_reporter():
+    from pycobertura.reporters import PhorgeJsonReporter
+
+    cobertura = make_cobertura('tests/dummy.source1/coverage.xml')
+    report = PhorgeJsonReporter(cobertura)
+
+    assert report.generate() == """\
+{
+    "dummy/__init__.py": "",
+    "dummy/dummy.py": "CCNCUU",
```
aconrad, line 824: What is N?

pycobertura/reporters.py
```diff
@@ -174,6 +174,28 @@ def generate(self):
...
+class PhorgeJsonReporter(Reporter):
+    """
+    A reporter that outputs a JSON object in the Phorge/Phabricator
+    coverage format. It is required to have the source files available.
+    """
+
+    def generate_one(self, filename):
+        statuses = self.cobertura.line_statuses(filename)
+        with self.cobertura.filesystem.open(filename) as f:
+            lines = ["N" for _ in f]
```
aconrad, line 186: Could we do something like `["N"] * len(f)`? On my phone now so I can't test.

pycobertura/reporters.py
```diff
@@ -174,6 +174,28 @@ def generate(self):
...
+            for lineNum, covered in statuses:
+                lines[lineNum - 1] = "C" if covered else "U"
```
aconrad, line 188: I have a long overdue PR that will introduce partially covered lines. Is there a character you'd like to use when I merge this PR eventually?

aconrad (commented):
Thanks for this! I'll take a closer look this week when I'm at my laptop.

tests/test_reporters.py
tusooa: This is described in https://we.phorge.it/book/phorge/article/arcanist_coverage/#building-coverage-support (I have put this link into README): N means "not executable"; C means "covered" and "U" means "not covered"

pycobertura/reporters.py
tusooa: As per phorge docs, it does not support partially covered lines.

pycobertura/reporters.py
tusooa: I don't think so, as the file handle is not an array at all.

```
] python -c 'with open("setup.cfg") as f: print(len(f))'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: object of type '_io.TextIOWrapper' has no len()

] python -c 'with open("setup.cfg") as f: print(len(["N" for _ in f]))'
50
```

PR #195, opened 2025-06-15
Feature/show columns

gro1m:
Another attempt of implementing show columns based on https://github.com/aconrad/pycobertura/pull/160.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ htmlcov/
 .tox/
 .coverage
 .cache
+.history
 nosetests.xml
 coverage.xml
 junit.xml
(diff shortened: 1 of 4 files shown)


gro1m:
This is why I only did for the `show` and not for the `diff` command yet and also no tests for the moment in order to get your feedback on priority.

aconrad:
@gro1m thanks so much for looking into this, I really appreciate it. I think it could be useful to have this feature since it was requested. Although honestly I don't know how common it is of a request. There aren't many columns to begin with.

But if we were to implement this in the least intrusive way, could we just pass the list of columns to render to the template? I understand it would be a waste of computation because we'd be gathering data for all columns and only rendering the ones we care at the end. But I feel it would be an easier implementation to begin with. What do you think?

PR #197, opened 2025-10-01
feat: add --fail-threshold option

paveltsialnou:
Add new `--fail-under` option that allows to return non-zero exit code when `pycobertura show cobertura.xml` is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,12 @@ $ pycobertura show --format github-annotation tests/cobertura.xml
 ::notice file=dummy/dummy4.py,line=1,endLine=6,title=pycobertura::not covered
 ```
 
+The following shows how to return a non-zero exit code when the line rate falls below the specified threshold.
+
+```shell
+$ pycobertura show --fail-under=80 cobertura.xml
+```
+
 If you run it in GitHub Actions/Apps, the above log generates check annotations.
 
 ![Example output of github-annotation formatted pycobertura show command](images/example_github_annotation_show.png)
--- a/README.md
+++ b/README.md
@@ -219,10 +219,10 @@ $ pycobertura show --format github-annotation tests/cobertura.xml
 ::notice file=dummy/dummy4.py,line=1,endLine=6,title=pycobertura::not covered
 ```
 
-The following shows how to return a non-zero exit code when the line rate falls below the specified threshold.
+The following shows how to return a non-zero exit code when the total number of uncovered lines falls above the specified threshold.
 
 ```shell
-$ pycobertura show --fail-under=80 cobertura.xml
+$ pycobertura show --fail-threshold=123 cobertura.xml
 ```
 
 If you run it in GitHub Actions/Apps, the above log generates check annotations.
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ $ pycobertura show --format github-annotation tests/cobertura.xml
 ::notice file=dummy/dummy4.py,line=1,endLine=6,title=pycobertura::not covered
 ```
 
-The following shows how to return a non-zero exit code when the total number of uncovered lines falls above the specified threshold.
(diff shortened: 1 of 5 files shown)


paveltsialnou:
@aconrad thanks for sharing your thoughts. I agree that in many cases that approach gives a more stable signal than percentage‑based checks, and your example highlights that well.

At the same time, I’ve seen a lot of teams that either aren’t ready to set up historical baselines for diffs or are required by policy to keep overall coverage above a set percentage.

The `--fail-under` flag here is entirely optional, so it doesn’t change the default behaviour, but it gives those teams a way to enforce a minimum % coverage gate while still being able to adopt diff‑coverage philosophy later.

I’m happy to make sure the docs clearly promote uncovered‑lines and diff as the stronger metric for regression prevention, and position percentage as a secondary tool for those who need it.

aconrad:
> At the same time, I’ve seen a lot of teams that either aren’t ready to set up historical baselines for diffs or are required by policy to keep overall coverage above a set percentage.

With the `diff` tool, the baseline is simply the target branch’s coverage. It tells you whether coverage has increased or decreased with the new changes (typically in a PR).

> The --fail-under flag here is entirely optional, so it doesn’t change the default behaviour

Absolutely, agreed.

> but it gives those teams a way to enforce a minimum % coverage gate while still being able to adopt diff-coverage philosophy later.

Let’s walk through a quick example.

Say you set `--fail-under` to 80%, and your codebase currently sits at 85%. Everything passes for now. Ideally, teams continue improving coverage to 86%, 87%, and so on. But if coverage drops to 84%, 83%, etc., the build won’t actually fail until it dips below 80%.

In practice, no one wants to wait for coverage to fall that far before getting feedback, right? So the next logical step is to automatically update the baseline whenever coverage changes — for instance, setting the new baseline to the last known 85%.

However, when coverage is constantly near the baseline (as happens when you auto-update it), it leads to more false positives — as explained in [this paragraph](https://github.com/aconrad/pycobertura/?tab=readme-ov-file#why-is-the-number-of-uncovered-lines-used-as-the-metric-to-check-if-code-coverage-worsened-rather-than-the-line-rate). You might see PRs fail at 84.98% even though there’s no meaningful coverage regression.

One possible middle ground could be similar to what you’re suggesting: keeping a "fail under" threshold, but based on the **number of uncovered lines** instead of a percentage. Whether that threshold is `80` (percent) or `18426` (uncovered lines) doesn’t really matter for CI purposes — it’s just a numeric gate that determines whether to fail the build. The uncovered-lines metric also works better with automatic baseline updates and avoids those false positives.

Finally, if coverage moves from 18,426 → 18,449 uncovered lines in a PR, it’s not immediately clear *where* coverage dropped — maybe a function stopped being called or dead code was introduced. That’s exactly why the `diff` command exists: to show, visually, which lines lost coverage compared to the target branch. But, as you said, it's a next step. I just wanted to illustrate the full circle.

So, what do you think about making `--fail-under` operate on the number of uncovered lines instead? Would that work for you and your team’s workflow?

README.md
```diff
@@ -210,6 +210,12 @@ $ pycobertura show --format github-annotation tests/cobertura.xml
 ::notice file=dummy/dummy4.py,line=1,endLine=6,title=pycobertura::not covered
 ```
 
+The following shows how to return a non-zero exit code when the total number of uncovered lines falls above the specified threshold.
```
aconrad:
```suggestion
The following shows how to return a non-zero exit code when the total number of uncovered lines exceeds the specified threshold.
```

aconrad:
Thanks for this. Can you add an entry in the CHANGES file please, under the "Unreleased" paragraph (follow other release note examples for formatting)

aconrad (approved):
Thanks!

PR #176, opened 2024-02-12
Fix total values when ignore regex used

danctorres:
**Expected Behavior**:
The report should display the total values (stmts, miss, and cover), calculated for all files excluding those specified by the ignore_regex.

**Actual Behavior**:
Previously, when using ignore_regex to exclude files, the report's total values were incorrect. The values remained the same whether the ignore_regex argument was passed or not.

**Steps to Reproduce**:
`~/pycobertura >  pycobertura show --ignore-regex ".*Main|.*ISorted" --format markdown tests/cobertura.xml`

**Report Before Fix**:
```
| Filename                 |   Stmts |   Miss | Cover   | Missing   |
|--------------------------|---------|--------|---------|-----------|
| search/BinarySearch.java |      12 |      1 | 91.67%  | 24        |
| search/LinearSearch.java |       7 |      2 | 71.43%  | 19-24     |
| TOTAL                    |      34 |      3 | 90.00%  |           |
```

**Changes**:
- Added ignore_regex as an input to the functions responsible for calculating total values.
- Introduced a test to ensure correct total value calculation when using ignore_regex.

**Result**:
After the fix, the total values are now calculated correctly, reflected by the following report, generated with the same command.

**Report After Fix**:
```
| Filename                 |   Stmts |   Miss | Cover   | Missing   |
|--------------------------|---------|--------|---------|-----------|
| search/BinarySearch.java |      12 |      1 | 91.67%  | 24        |
| search/LinearSearch.java |       7 |      2 | 71.43%  | 19-24     |
| TOTAL                    |      19 |      3 | 84.21%  |           |
```

PR #202, opened 2026-08-27
Apply --ignore-regex to the TOTAL row of diff reports

Eljees:
### The problem

`--ignore-regex` filters files out of the body of a `diff` report, but the
`TOTAL` row is still computed over every file, so the totals disagree with
the rows printed above them. Ignoring everything makes it obvious:

```
$ pycobertura diff --ignore-regex ".*" tests/dummy.source1/coverage.xml tests/dummy.source2/coverage.xml
Filename      Stmts    Miss  Cover    Missing
----------  -------  ------  -------  ---------
TOTAL            -1      -4  +31.06%
```

The table is empty, and `TOTAL` still reports the numbers for the files that
were supposed to be excluded.

With this change:

```
$ pycobertura diff --ignore-regex ".*" tests/dummy.source1/coverage.xml tests/dummy.source2/coverage.xml
Filename      Stmts    Miss  Cover     Missing
----------  -------  ------  --------  ---------
TOTAL             0       0  +100.00%
```

### Why this looks like a bug rather than intended behaviour

- `show` already behaves the other way: #176 threaded `ignore_regex` into the
  total calculations for `show` and was released in 3.3.1, and
  `test_text_report__with_ignore_regex` pins that behaviour. `diff` was left
  out of that change.
- The lower layer is already prepared for it: `Cobertura.line_rate()`,
  `total_statements()`, `total_misses()` and `total_hits()` all take
  `ignore_regex`. Only the `CoberturaDiff` counterparts do not.

### The change

`DeltaReporter.get_summary_lines()` calls `diff_total_statements()`,
`diff_total_misses()` and `diff_line_rate()` without passing
`self.ignore_regex`, and those three (plus the `_diff_attr` helper behind
them) had no parameter to receive it. This threads `ignore_regex` through.

Because every delta format goes through the same `get_summary_lines()`, this
fixes the TOTAL row for text, csv, markdown, json, yaml and html at once.

### Tests

One CLI test and two reporter tests. Without the production change they fail
on the numbers, not on the shape of the output:

```
E   AssertionError: assert '-1' == '0'
E   AssertionError: assert '-1' == '-3'
E   AssertionError: assert 'TOTAL  -1  -4  +31.06%' == 'TOTAL  0  0  +100.00%'
```

| | passed | failed |
|---|---|---|
| master | 187 | 0 |
| new tests, without the fix | 187 | 3 |
| new tests, with the fix | 190 | 0 |

I also checked that both halves of the change are load-bearing by breaking
each one on its own and confirming the same three tests go red: reverting the
three call sites in `reporters.py` gives 187/3, and reverting only the file
selection in `_diff_attr` gives 187/3 as well.

`flake8 pycobertura/` and `black --check --diff pycobertura/` (the two tox
gates) both pass.


Eljees:
A note for whenever this gets a look: CI has never run on this PR, and that is not a failure on my side.

The check suite GitHub created for the `pycobertura` workflow is at `action_required` with zero runs in it, which is the "approve workflows for this contributor" gate. So the PR shows no test results simply because nothing has been allowed to start; a maintainer pressing "Approve and run" would produce real ones.

